PR #35, opened 2016-05-18
Fix TLS handshake error

bole5:
When connecting with Tunnelblick client I received following error:
```
TLS_ERROR: BIO read tls_read_plaintext error: error:140760FC:SSL routines:SSL23_GET_CLIENT_HELLO:unknown protocol
TLS Error: TLS object -> incoming plaintext read error
TLS Error: TLS handshake failed
```

To fix the problem the same **tls-version-min** directive used in server configuration has to be added to the client config.

### DIFF
--- a/Default.txt
+++ b/Default.txt
@@ -8,6 +8,7 @@ persist-key
 persist-tun
 key-direction 1
 remote-cert-tls server
+tls-version-min 1.2
 verify-x509-name SRVRNAME name
 cipher AES-256-CBC
 auth SHA256


0-kaladin:
hmm, i had this in there and suspect when i reworked the OS detection it got lost somewhere.  the installer added it to the file i believe.  Let me figure out what happened in the installer before i just add it to the default.txt

0-kaladin:
yea i never had this in the default.txt.  my tunnelblick didn't complain about it but i merged.  we'll see if other clients complain.

0-kaladin:
@bole5 what version and platform tunnelblick do you use?  As I stated before my tunnelblick never complained and I"m using 3.6.3 on macosx.  And my android openvpn client is telling me it is ignoring tls-version-min in client config as its unused.  I don't think it is a valid directive client side, server only.  
So i want to be sure it isn't something else that fixed your issue as i'm inclined to remove this patch.

bole5:
I am running Tunnelblick 3.5.3 (build 4270.4371) running openvpn 2.3.6. This directive only works with newer clients.

For example, my test setup using OpenVZ and debian7 is running openvpn 2.2.1 and there I had to remove both: `tls-version-min 1.2` (and tis_auth) and `verify-x509-name server name` in order to get connection working...

0-kaladin:
your tunnelblick is old.  3.6.3 is newer than 3.5.3.  The android client i'm using is written by arne schwab who directly contributes to openvpn, it is the latest version and doesn't recognize tls-version-min as a client directive.  I know older servers don't recognize it but I didn't care about that since pivpn is the server and I know what version it brings (which support tls-version-min).  So your test setup, running a server version 2.2.1 is an invalid test for anything pivpn.  
when you talk about the tunnelblick you have at 3.5.3, that tunnelblick upgraded to the latest version should connect fine to a pivpn installed server.  

now you didn't specify, but did you use this script to install on that openvz debian7 box?  if so, then still my fix would be to install a newer openvpn like i do on ubuntu rather than from the old debian repos.  
see if i detect ubuntu I do this:
```
if [[ $PLAT == "Ubuntu" ]]; then
        wget -O - https://swupdate.openvpn.net/repos/repo-public.gpg| $SUDO apt-key add -
        echo "deb http://swupdate.openvpn.net/apt trusty main" | $SUDO tee /etc/apt/sources.list.d/swupdate.openvpn.net.list > /dev/null
    fi
```
what i might need to do is modify that for debian as well.  

but this is why i state in the readme, etc that it's supported on raspbian jessie and ubuntu 14.04 since there could be all kinds of little things for other releases.  

so if all this is the case, (you used script to install on debian7, please send me the output of
```
lsb_release -si
lsb_release -sc
```

you did get a popup during install stating your OS was *maybe* supported, correct?

bole5:
Maybe I was unclear in my reply, I installed pivpn server on my new Raspberry Pi 3 running jessie as in the instructions. The problem is due to using different openvpn clients. My TunnelBlick is not able to connect without the tls-auth directive in the client config file.

The reference to the OpenVZ and debian7 was just for your information that the client directive `verify-x509-name server name` also creates error on older openvpn **clients** as I used a config generated by pivpn.

Since pivpn is configuring server to run with tls-auth, for a good reason as this is the safest option, clients that don't support this option will not be able to connect. You might want to add that information to the readme and display a warning when generating new client configs.

Alternatively you could ask users whether or not they want to use tis-auth during the setup process and create configuration based on user reply.

0-kaladin:
clients typically don't stay old though.  why didn't you upgrade your tunnelblick?

PR #91, opened 2016-09-28
Feature/add tcp option

StephenKinger:
I have add a new menu for tcp option instead of udp which can be useful for users who needs to use 443 port for example in a company to connect to an external vpn.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -476,13 +476,11 @@ update_repo() {
 }
 
 setCustomProto() {
-  # Turn the available interfaces into an array so it can be used with a whiptail dialog
+  # Set the available protocols into an array so it can be used with a whiptail dialog
   protoArray=()
-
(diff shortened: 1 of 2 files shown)


auto_install/install.sh
```diff
@@ -475,6 +475,32 @@ update_repo() {
     echo " done!"
 }
 
+setCustomProto() {
```
StephenKinger: yes, we can add a message to say that UDP is recommanded

redfast00: @StephenKinger @0-kaladin: added the message and some cleanup. Should be good to go

StephenKinger:
Ok, i checked and it's work. I can set up automatically the 443 port in case of tcp is choosen (useful for vpn setup used to be accessed from filtered network) if you want.

0-kaladin:
Looks good guys, great work!

PR #94, opened 2016-10-04
Added option to download dh parameters

redfast00:
I added this because generating DH parameters (especially 4096 bits) on a RPi takes a really long time (I let it run for 12 hours and it still hasn't finished). The keys are downloaded from [2ton.com.au](https://2ton.com.au/dhtool/#service).

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -715,9 +715,9 @@ confOpenVPN() {
     # Build the server
     ${SUDOE} ./build-key-server --batch $SERVER_NAME
 
-    if (whiptail --backtitle "Setup OpenVPN" --title "Diffie-Hellman Parameters" --yesno "Generating Diffie-Hellman might take a long time on a Raspberry Pi. Do you want to download them?" $r $c)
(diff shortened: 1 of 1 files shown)


0-kaladin:
Didn't know you could download.  I feel like there is some security issue downloading them vs having them generated.  Maybe not.  Could we make this an option only if key size is 4096?  Then it makes sense to offer this but with a smaller key size it isn't that bad to hit some random .au site to grab it.  
Just keep in mind the mindset we need to be in here.  This script is meant to lower the bar of entry for anyone to get their own openvpn server up and running.  that means there will be a ton of users who don't even know what a key is or whether they really care about the size.  Hitting enter through this script should give them our best recommendations at all times without asking them questions they may not know the answer to (to the best of our ability).  
I greatly appreciate the contributions I'm just saying this to ensure we stay focused. 
Thanks!

redfast00:
Indeed, Diffie Hellman parameters can be backdoored, so I added a short warning (If you're paranoid, choose "No") and the option to download is now only available if keysize >= 4096 bits.

redfast00:
I messed up my git... Closing this and reopening later. :fire:

PR #95, opened 2016-10-04
Sanitization 'n input validation

redfast00:
Now checks if password is not blank, checks if password length is between 4 and 1024 characters, that the name doesn't contain any special characters (like spaces) and cleaned up source.  
I might add some more commits later


0-kaladin:
Some good cleanup here, thanks!

Issue #28, opened 2016-05-08
Error Creating client crt


airforceboricua:
Getting this when running 'pivpn add':

```
Enter a Name for the Client:  test
NOTE: If you run ./clean-all, I will be doing a rm -rf on /etc/openvpn/easy-rsa/keys
/etc/openvpn/easy-rsa/vars: line 65: unexpected EOF while looking for matching ``'
/etc/openvpn/easy-rsa/vars: line 81: syntax error: unexpected end of file
Enter the password for the Client:
Enter the password again to verify:
spawn ./build-key-pass test
pkitool: Need a readable ca.crt and ca.key in /etc/openvpn/easy-rsa/keys
Try pkitool --initca to build a root certificate/key.
expect: spawn id exp6 not open
    while executing
"expect "Verifying - Enter PEM pass phrase" { send "test123\r" }"
spawn openssl rsa -in test.key -des3 -out test.3des.key
Error opening Private Key test.key
1995716704:error:02001002:system library:fopen:No such file or directory:bss_file.c:398:fopen('test.key','r')
1995716704:error:20074002:BIO routines:FILE_CTRL:system lib:bss_file.c:400:
unable to load Private Key
expect: spawn id exp6 not open
    while executing
"expect "Enter PEM pass phrase" { send "test123\r" }"
[ERROR]: Client Public Key Certificate not found: test.crt
```

Any ideas?  Thanks for all of this by the way.  Made everything very simple.

0-kaladin:
hmm can you 'cat /etc/openvpn/easy-rsa/vars'
and paste that output here?

airforceboricua:
Sure can.

`# easy-rsa parameter settings

# NOTE: If you installed from an RPM,
# don't edit this file in place in
# /usr/share/openvpn/easy-rsa --
# instead, you should copy the whole
# easy-rsa directory to another location
# (such as /etc/openvpn) so that your
# edits will not be wiped out by a future
# OpenVPN package upgrade.

# This variable should point to
# the top level of the easy-rsa
# tree.
export EASY_RSA="/etc/openvpn/easy-rsa"

#
# This variable should point to
# the requested executables
#
export OPENSSL="openssl"
export PKCS11TOOL="pkcs11-tool"
export GREP="grep"


# This variable should point to
# the openssl.cnf file included
# with easy-rsa.
export KEY_CONFIG=`$EASY_RSA/whichopensslcnf $EASY_RSA`

# Edit this variable to point to
[...]

Issue #99, opened 2016-10-08
"unable to load Private Key"


cmsimike:
Hi,

Recently, whenever I try to add a new client, I am getting this exception:

```
spawn openssl rsa -in iphone7plus.key -des3 -out iphone7plus.3des.key
unable to load Private Key
1995843680:error:0906D06C:PEM routines:PEM_read_bio:no start line:pem_lib.c:696:Expecting: ANY PRIVATE KEY
expect: spawn id exp6 not open
    while executing
"expect "Enter PEM pass phrase" { send "\r" }"
[ERROR]: Client Public Key Certificate not found: iphone7plus.crt

```

This happened on an already existing install and (what I believe to be) a clean reinstall of pivpn.
Any ideas?

redfast00:
@0-kaladin: this might have something to do with #28 #62 #85. Investigating...

redfast00:
@cmsimike: can you please post the result of `cat /etc/openvpn/easy-rsa/vars`?

cmsimike:
Here are the contents of the file:
```
# easy-rsa parameter settings

# NOTE: If you installed from an RPM,
# don't edit this file in place in
# /usr/share/openvpn/easy-rsa --
# instead, you should copy the whole
# easy-rsa directory to another location
# (such as /etc/openvpn) so that your
# edits will not be wiped out by a future
# OpenVPN package upgrade.

# This variable should point to
# the top level of the easy-rsa
# tree.
export EASY_RSA="/etc/openvpn/easy-rsa"

#
# This variable should point to
# the requested executables
#
export OPENSSL="openssl"
export PKCS11TOOL="pkcs11-tool"
export GREP="grep"


# This variable should point to
# the openssl.cnf file included
# with easy-rsa.
export KEY_CONFIG=`$EASY_RSA/whichopensslcnf $EASY_RSA`

# Edit this variable to point to
# your soon-to-be-created key
# directory.
#
# WARNING: clean-all will do
# a rm -rf on this directory
# so make sure you define
# it correctly!
export KEY_DIR="$EASY_RSA/keys"

# Issue rm -rf warning
echo NOTE: If you run ./clean-all, I will be doing a rm -rf on $KEY_DIR

# PKCS11 fixes
export PKCS11_MODULE_PATH="dummy"
export PKCS11_PIN="dummy"

# Increase this to 2048 if you
# are paranoid.  This will slow
# down TLS negotiation performance
# as well as the one-time DH parms
# generation process.
[...]

Issue #85, opened 2016-09-22
Pem password when adding vpn user


kingcatdaan:
i get this error no matter what i do

root@Anna:~# pivpn add
Enter a Name for the Client:  daan
NOTE: If you run ./clean-all, I will be doing a rm -rf on /etc/openvpn/easy-rsa/keys
Enter the password for the Client:
Enter the password again to verify:
spawn ./build-key-pass daan
Generating a 4096 bit RSA private key
....................................................................................................................++
......................................................................++
writing new private key to 'daan.key'
Enter PEM pass phrase:
spawn openssl rsa -in daan.key -des3 -out daan.3des.key
unable to load Private Key
3069334736:error:0906D06C:PEM routines:PEM_read_bio:no start line:pem_lib.c:696:Expecting: ANY PRIVATE KEY
expect: spawn id exp6 not open
    while executing
"expect "Enter PEM pass phrase" { send "passwd\r" }"
[ERROR]: Client Public Key Certificate not found: daan.crt

ant1612:
Hi,

I've got this error before. The only way I've "fixed" this is to repeat the same command again. Seems to work the second time round. 

I've done this for two/three different users and each time it appears with a fault

After the second attempt on each user, I've not issues logging in or it appearing to be at fault

michielvanhoorn:
I have this issue too but doing it twice does not help.

BOZG:
It worked fine for me when I first installed PiVPN. I need to uninstall and then reinstall and after that, this error kept appearing.

Issue #62, opened 2016-08-02
generating profile fails


matthijsl:
```
spawn ./build-key-pass client
Generating a 4096 bit RSA private key
............................................................................................................................++
................................++
writing new private key to 'client.key'
Enter PEM pass phrase:myphrase
spawn openssl rsa -in client.key -des3 -out client.3des.key
unable to load Private Key
1996052576:error:0906D06C:PEM routines:PEM_read_bio:no start line:pem_lib.c:696:Expecting: ANY PRIVATE KEY
expect: spawn id exp6 not open
    while executing
"expect "Enter PEM pass phrase" { send "theclientpassword\r" }"
Client's cert found: client.crt
[ERROR]: Client 3des Private Key not found: client.3des.key
root@raspberrypi:/etc/openvpn/easy-rsa/keys# myphrase
bash: myphrase: command not found

```

I'm running into this while trying to `pivpn add` a profile.

running `pivpn add [nopass]` works but has me ending up with tunnelblick complaining about an unaccepted passphrase

MatthijsLiethof:
I've been getting very intermittent results, following trying only letters instead of alphanumeric passwords. And even then with short passwords I've needed to run the command with the same input over and over just to get 1 success.

MatthijsLiethof:
I succeeded in getting a working profile with certificates included. I did this by breaking down the steps found in `makeOVPN.sh` what I found is that with the private keys generated through `pivpn -a` fail in the step where you create the des3 file https://github.com/pivpn/pivpn/blob/master/scripts/makeOVPN.sh#L82

Getting an error saying that it was expecting a private key file. 

When I ran `./build-key-pass` manually I was able to step through with an alphanumeric passphrase of significant length. Running these keys through the `openssl rsa -in $NAME$OKEY -des3 -out $NAME$KEY` command did not fail.

After that I used a makeshift script to create the profile.
```
#!/bin/bash
CLIENT_FILE="client.ovpn"
CLIENT_NAME="client"

touch $CLIENT_FILE
cat Default.txt >> $CLIENT_FILE

# Echo the needed certificates 
echo '<ca>' >> /etc/openvpn/easy-rsa/keys/$CLIENT_FILE
cat /etc/openvpn/easy-rsa/keys/ca.crt >> /etc/openvpn/easy-rsa/keys/$CLIENT_FILE
echo '</ca>' >> /etc/openvpn/easy-rsa/keys/$CLIENT_FILE

[...]

PR #102, opened 2016-10-09
fixes expect timeout

redfast00:
Fixes: #28 #62 #85 #99. This was caused by `expect` timing out if the key generation took too long.

### DIFF
--- a/scripts/makeOVPN.sh
+++ b/scripts/makeOVPN.sh
@@ -20,6 +20,7 @@ function keynoPASS() {
 
     #Build the client key
     expect << EOF
+    set timeout -1
     spawn ./build-key "$NAME"
     expect "Country Name" { send "\r" }
     expect "State or Province Name" { send "\r" }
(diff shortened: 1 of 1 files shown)


0-kaladin:
ah, good catch!

mixpc:
Beautiful! Please, kindly let me know the linux command to update pivpn and test it. I've tried
```
sudo apt-get update
sudo apt-get upgrade
```
but is seems there is no update yet.

```
me@mydevice:~ $ sudo apt-get -u upgrade
Reading package lists... Done
Building dependency tree
Reading state information... Done
Calculating upgrade... Done
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
me@mydevice:~ $
```

BOZG:
@mixpc Unless things have changed, the only supported method to upgrade PiVPN is to [uninstall and reinstall.](https://github.com/pivpn/pivpn/wiki/Upgrading)

PR #111, opened 2016-10-15
Cleanup

redfast00:
I edited the bash-completion so it doesn't suggest two commands (like `pivpn add <tab>` suggested `add debug help list revoke uninstall`, now only suggests `nopass`)

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -12,9 +12,6 @@
 
 ######## VARIABLES #########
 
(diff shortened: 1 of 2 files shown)


redfast00:
@peakyc @0-kaladin @StarshipEngineer Can someone please do a review of this? :smiley:

redfast00:
Okay, I manually tested this with the Dietpi VirtualBox image. Should be good to go! :smiley:

PR #128, opened 2016-10-20
Randomize DH params fetched from 2ton.com.au

shaungrady:
Fetch random DH params from the last 128 generated by 2ton.com.au. This will reduce the likelihood that multiple installs in short period of time will share the same DH params.


redfast00:
Awesome! Thank you

PR #129, opened 2016-10-21
Explain DH param download in greater detail

shaungrady:
New text:

> Download Diffie-Hellman parameters from a public DH parameter generation service?
> 
> Generating DH parameters for a $ENCRYPT-bit key can take many hours on a Raspberry Pi. You can instead download DH parameters from "2 Ton Digital" that are generated at regular intervals as part of a public service. Downloaded DH parameters will be randomly selected from a pool of the last 128 generated.
> More information about this service can be found here: https://2ton.com.au/dhtool/
>
> If you're paranoid, choose 'No' and Diffie-Hellman parameters will be generated on your device.

- Change default from "no" to "yes," since it's perfectly safe for 99% of users **(Reverted to "no" being default)**
- Be a bit more specific about what "a long time" means
- Provide a bit of information about the service
- Provide a link to read more about the service

Let me know if you'd any of the wording changed.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -743,7 +743,7 @@ confOpenVPN() {
     # Build the server
     ${SUDOE} ./build-key-server --batch $SERVER_NAME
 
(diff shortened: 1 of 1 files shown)


redfast00:
Hi, thanks for contributing. I previously discussed this with @0-kaladin (the maintainer of PiVPN) and they thought it should default to 'no'. The rest looks good.

shaungrady:
No problem! Done 👍

redfast00:
Awesome! :+1:

PR #165, opened 2016-12-08
Some fixes/hacks for RaspPi wheezy

bglnelissen:
PRETTY_NAME="Raspbian GNU/Linux 7 (wheezy)"
NAME="Raspbian GNU/Linux"
VERSION_ID="7"
VERSION="7 (wheezy)"
ID=raspbian
ID_LIKE=debian

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -911,8 +911,12 @@ confOVPN() {
     # verify server name to strengthen security
     $SUDO sed -i "s/SRVRNAME/${SERVER_NAME}/" /etc/openvpn/easy-rsa/pki/Default.txt
 
-    $SUDO mkdir "/home/$pivpnUser/ovpns"
-    $SUDO chmod 0777 -R "/home/$pivpnUser/ovpns"
+  # Here we expect the user (${pivpnUser}) to have it's own home dir...
+  # Some users do not have an own homedir. ("mkdir: cannot create directory `/home/XXXX/ovpns': No such file or directory")
+  # We should ask the user were to put the 'ovpns' dir.
+  # For now I hacked it using a '-p' flag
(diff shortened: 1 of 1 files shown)


0-kaladin:
First, thanks for contributing! Def could use more eyes and more people testing esp on stuff I don't test, like wheezy.  
So a few comments:
your 'fixed wget command' https://github.com/pivpn/pivpn/pull/165/commits/b632e69cdf18f02e7f7cdd9ec252316e767eddea
is safe to merge.  my code didn't complain to me but I agree it is incorrect, my only thought is maybe the older wget in wheezy is less tolerant of errors?

then your fixed easyrsa install commit which results in bad dir you fixed with the fixed download location commit.  So I'd like to add that as well.

The homedir one I do not want.  Here's why:
The vast majority have a user account with a home dir.  by default raspian comes with one.  The default process to create any user will also create the users home dir.  if they have a user without a home dir then chances are that user can't login (it may have a different case).  Regardless the only real purpose of dropping ovpn files there is for easy user file transfer which won't be that easy if user has no home dir.  What I would prefer to do for this edge case is actually only show users with home dirs.  if it doesn't see any then prompt user to create one (with a home dir) which it will do today if it detects no normal users.  so for now, cause i don't have time to implement that, can you drop this commit?

Also I'd love if you can submit this to the test branch.  Basically fork from my test branch, give me the 'fixed wget command' and the 'fixed download location' change and I can merge them into test. 

then we can test install from the test branch to ensure nothing broke with
`curl -L http://install-test.pivpn.io | TESTING=1 bash`

0-kaladin:
I don't have much time and wanted to get this into the test branch for testing so you'll see i was able to cherry-pick what i wanted from you in there.  If you can confirm an install from test branch works for you then I will merge to master.  (except the home dir thing)

Issue #232, opened 2017-03-30
change settings in setup


RemcoMulder2000:
Hi all,

I try to install pivpn using TCP and testlevel encryption (1024) instead op UDP and 2048 bit encryption. In the setup I select the settings I want and then hit enter but it still uses the UDP and 2048 bit settings.

Enyone who knows an solution to this?

cfcolaco:
Just finished running piVPN with the settings you mentioned and piVPN applied the options correctly. 

proto tcp
port 443
dh /etc/openvpn/easy-rsa/pi/dh1024.pem

probably you missed something during the setup
the easier solution I believe would be reinstall piVPN and check where you missed.
People usually highlight the new setting and don't press "space bar" to actually select it, could it be the case?

otherwise, and if I am not mistaken, you will have to generate all the certificates and encryption keys manually and place the correct path to the certificates in the the /etc/openvpn/server.conf file which is prone to harder to track mistakes. 

Hope this helps. if you try reinstalling please confirm back to us if it solved. 
thank you

SloCompTech:
Hi, I have posibly same problem, I selected 4096bit key mode, but in setup shows using 2048bit key

cfcolaco:
@SloCompTech  are you sure you really selected the right option??? 

Will close the issue, however, feel free to reopen it or ask to reopen it if needed! will be glad to help out!

PR #243, opened 2017-04-17
Clearly and consistently indicate that a user selects a list item with the space bar.

jellemdekker:
During the installation procedure, a user is presented multiple times with a list of items to choose from. These are selected by pressing the space bar. However, this is only indicated textually at the step where the user selects which interface to use. This pull request adds this text to every item selection step. Presumably fixes issue #232 .

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -173,7 +173,7 @@ chooseUser() {
         fi
         userArray+=("${line}" "" "${mode}")
     done <<< "${availableUsers}"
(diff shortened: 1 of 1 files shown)


redfast00:
Looks fine to me, thanks for the PR :D waiting for @0-kaladin to merge.

redfast00:
Ping @0-kaladin

0-kaladin:
Will look. Thx

PR #277, opened 2017-06-16
dynamic pivpn user home

orttez:
Let's take into consideration the posibility that pivpn user won't be always in /home, and the use of other locations like /usr/local is possible.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1005,7 +1005,7 @@ confOVPN() {
     $SUDO sed -i "s/SRVRNAME/${SERVER_NAME}/" /etc/openvpn/easy-rsa/pki/Default.txt
 
     $SUDO mkdir "$pivpnHome/ovpns"
-    $SUDO chmod 0777 -R "$pivpnHome/ovpns"
+    $SUDO chmod 0755 -R "$pivpnHome/ovpns"
 }
 
 installPiVPN() {
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -186,6 +186,7 @@ chooseUser() {
     if [[ $? = 0 ]]; then
         for desiredUser in ${chooseUserOptions}; do
             pivpnUser=${desiredUser}
+				pivpnHome=$(eval echo ~${desiredUser})
             echo "::: Using User: $pivpnUser"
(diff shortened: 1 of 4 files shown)


cfcolaco:
Nice work, and well thought, however, I don't consider it a good option for a few reasons, most of them related to security.

1 - OVPN files are very sensitive files, so It is never a good idea to give them chmod 777 as you are doing in line 108 of install.sh, it should be at the most 660. 

2 - It's well known and supposed to be common sense, that no one should ssh as root, having this said, it's recommended to have at least one unprivileged user, to access via ssh and then scale privileges either using sudo or su to become root, so the standard place for its home should be /home/$USER so makes sense to give the OVPNS to a user and only that user and root could reach those files, no one else on the system.

What I consider to be some good solutions to overcome the issue of users giving the wrong user to hold the OVPN files rather than supporting a dynamic home directory: 

1 - detect available users, and check if there's a home directory for that user before presenting it to a user to choose. 

2 - Create PiVPN user with password to hold the files, which eliminates the need for any user installing to choose a user and providing the wrong one, which makes a litle bit more secure to keep the OVPN files. 

3 - Place the OVPN files in a fixed directory, and only a privileged user ( root or sudo ) can access to them, which can be a bit more problematic to overcome the file permissions in this case. 


don't take me wrong, you did a good job, this is just my considerations, about it.
@redfast00 and @0-kaladin have the word here! ;)

orttez:
@cfcolaco Thanks for the feedback.
I updated my branch with /home/$USER permissions I consider more suitable, but the main repo will need that as well.

Just to through my 0011 cents, I like your suggestion. I would cross out option 1 and go straight to option 2. The PiVPN user will be created anyways if it doesn't exist, so I'd get rid of the username prompt.

cfcolaco:
@orttez wait for @redfast00 or @0-kaladin come around here they might have a different opinion or something already in mind for that! =)

cfcolaco:
@redfast00 @0-kaladin what about this PR??

cfcolaco:
@redfast00 @0-kaladin shall we close this PR.. and open a ticket to find a better solution for this??

redfast00:
Yeah, sure. I think your suggestions are quite good, so please copy them over in the issue as well.

PR #284, opened 2017-06-25
Added ability to disable logging on the server during installation

jellemdekker:
I added a new selection menu where the user can specify whether or not to disable logging on the OpenVPN server (e.g. redirect log output to `/dev/null`). Logging is still enabled by default, but can now be disabled if desired for better privacy. I also corrected a typo in the variable names that contain the location of the installation logfile. 

Note how I'm using `#` instead of `/` as the separator-character in sed. This eliminates the need for escaping paths to files and folders, and improves the overall readability of the expression. I think this might be useful for the other expressions as well.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -799,7 +799,7 @@ setClientDNS() {
 setLoggingPolicy() {
     CONLOGLOC="/var/log/openvpn.log"
     STATUSLOGLOC="/var/log/openvpn-status.log"
-    if (whiptail --backtitle "Logging policy" --title "Disable logging" --yesno --defaultno "Do you want to disable logging of connection information?" ${r} ${c}) then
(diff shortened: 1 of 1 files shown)


redfast00:
This looks good. Note that if someone is able to read the logs, they can probably do much worse stuff (for example MITM the VPN). This is still useful in case the server gets compromised (so itt doesn't leak who connected at what time). I am however not really sure if we should offer this option: it is an extra hurdle for any (non-tech) users trying to install this. Any input @0-kaladin ?

jellemdekker:
Please tell me if my PR is useful for your project.

redfast00:
I haven't heard from @0-kaladin for quite some time now, so I don't know what he thinks about this. I (that's just me though) don't think this is worth the additional hurdle seen what gets leaked if an attacker would compromise the server. Thanks for the PR tho! Btw, if you're looking to contribute to PiVPN, we'd love to get support for Raspbian Stetch.

Issue #280, opened 2017-06-21
Update to OpenVPN 2.4.3 or 2.3.17


AdySan:
# PiVPN Issue Template

## Console output of `pivpn debug`
```
::: Generating Debug Output
:::                                     :::
::              PiVPN Debug              ::
:::                                     :::
::      Latest Commit                    ::
:::                                     :::
commit 6ce39bfec399d9a8203409a108ef27f2ee29327a
Author: 0-kaladin <0.kaladin@gmail.com>
Date:   Mon Feb 27 21:01:23 2017 -0500

    Update README for IRC channel

    Due to discontinuation of Google Spaces
:::                                     :::
::      Recursive list of files in       ::
::      /etc/openvpn/easy-rsa/pki        ::
:::                                     :::
/etc/openvpn/easy-rsa/pki/:
AdyPiOpenVPN.ovpn
ca.crt
Default.txt
dh2048.pem
index.txt
index.txt.attr
index.txt.attr.old
index.txt.old
issued
private
serial
serial.old
ta.key

/etc/openvpn/easy-rsa/pki/issued:
AdyPiOpenVPN.crt
server.crt

/etc/openvpn/easy-rsa/pki/private:
AdyPiOpenVPN.key
ca.key
server.key
:::                                     :::
::      Output of /etc/pivpn/*           ::
:::                                     :::
:: START /etc/pivpn/DET_PLATFORM ::
Raspbian
:: END /etc/pivpn/DET_PLATFORM ::
:: START /etc/pivpn/INSTALL_PORT ::
1194
:: END /etc/pivpn/INSTALL_PORT ::
:: START /etc/pivpn/INSTALL_PROTO ::
udp
:: END /etc/pivpn/INSTALL_PROTO ::
:: START /etc/pivpn/INSTALL_USER ::
pi
:: END /etc/pivpn/INSTALL_USER ::
:: START /etc/pivpn/NO_UFW ::
1
:: END /etc/pivpn/NO_UFW ::
:: START /etc/pivpn/pivpnINTERFACE ::
eth0
:: END /etc/pivpn/pivpnINTERFACE ::
:: START /etc/pivpn/REVOKE_STATUS ::
0
:: END /etc/pivpn/REVOKE_STATUS ::
:::                                     :::
:: /etc/openvpn/easy-rsa/pki/Default.txt ::
:::                                     :::
client
dev tun
proto tcp
remote adysan.ddns.net 22
resolv-retry infinite
nobind
persist-key
[...]

PR #292, opened 2017-06-29
changed to new openvpn repos

cfcolaco:
Changed the OpenVPN repositories to the new ones, 

It will use OpenVPN stable repo and is now installing version 2.4.3

this solves issue #280 

tested the script everything worked well!


redfast00:
Looks good, thanks!

PR #325, opened 2017-08-05
Added Related Projects section

InnovativeInventor:
This is mainly to mention the docker-pivpn project. I wasn't sure what to do about the Contributions section because it does contain a few of the projects in the Related Projects section, so I just left it alone.


redfast00:
Thanks!

PR #361, opened 2017-09-23
Generate and activate CRL during installation

jellemdekker:
Implemented that an empty certificate revocation list is generated during installation after generation of other parts of the Public Key Infrastructure. The added benefit of this is that whenever the user now revokes a client, the change is instant. Whereas before, the first time a revocation happened since the installation of PiVPN, the OpenVPN server had to be restarted to enable the then-newly-generated CRL (causing an interruption of service for other connected clients). This change also makes the file /etc/pivpn/REVOKE_STATUS obsolete.

Documentation: https://openvpn.net/index.php/open-source/documentation/howto.html#revoke

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -886,8 +886,8 @@ EOF
 
     # Generate an empty Certificate Revocation List
     ${SUDOE} ./easyrsa gen-crl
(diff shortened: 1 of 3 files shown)


jellemdekker:
Please hold off on merging this, I'm investigating a problem with permissions on the CRL file.

PR #375, opened 2017-10-09
Add human readable values for bytes received/sent

azlux:
Code to add human readable value for `pivpn clients`
The bytes value was a little bit useless in this form.
The code check if the function `numfmt` exist into the shell context to avoid error.

Az


redfast00:
This looks good, thanks! Have you tested this?

azlux:
Yes, every day.
I just don't know if you want 1024bytes as 1kib or 1 kb

PR #378, opened 2017-10-12
[Needs testing] Extended the expiration date of the certificate revocation list to 10 years.

jellemdekker:
The default expiration date for the certificate revocation list was 180 days after the previous revocation (or since the installation of PiVPN). If a user doesn't revoke a profile every 180 days, then clients will not be able to connect after 180 days. Therefore, I have extended the expiration date of the certificate revocation list to 10 years.


jellemdekker:
This PR still needs to be tested, because of the usage of `{SUDOE}` on line 889 in `auto_install/install.sh`. I will get to this later this week.

jellemdekker:
Currently waiting on feedback on the issue referenced above

PR #379, opened 2017-10-12
Removed excess code

jellemdekker:
Commented out a number of excess options that were set implicitly by using the `server` option. Source: [OpenVPN 2.4 man page (--server directive)](https://community.openvpn.net/openvpn/wiki/Openvpn24ManPage#lbAH)

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -367,16 +367,6 @@ setStaticIPv4() {
     fi
 }
 
-#setNetwork() {
-    # Sets the Network IP and Mask correctly
-    #export PATH=${PATH}:/sbin:/usr/sbin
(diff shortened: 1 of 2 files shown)


jellemdekker:
I have tested this and here are the results:
- I can connect with multiple clients at the same time,
- They all receive a 10.8.0.X IP address
- I can access IP addresses (192.168.X.X) on the remote LAN.
- I can access the internet

Can anyone test if this behavior is the same for them or if they are experiencing problems?

orazioedoardo:
Can you post the output of `route -n` from a client both with and without the change?

jellemdekker:
Great, thank you for your feedback. This pull request is now ready to be reviewed and merged by a collaborator. @redfast00 @0-kaladin

PR #383, opened 2017-10-17
fixed pivpn clients text display

cfcolaco:
Fixes titles in pivpn clients not being properly aligned. 
Fixes connected users information that was being displayed on a single line instead of one line per connected client. 

making PR Directly into master as it is a minor fix to clientStat.sh


cfcolaco:
hmm ... weird .. looking at the conflicts seems my local repos are not fetching the correct info from the upstream! =S So I don't have the pr #375 ... weird ... closing and fixing the conflict ... making pr with the fixed conflict.

PR #384, opened 2017-10-18
Fix Clients only in one line

cfcolaco:
pivpn clients script was showing clients in only one line, and titles were not properly aligned. 

Fixed confits due to my local repo not syncing correctly, tested, everything is fine now!


redfast00:
Thanks!

Issue #364, opened 2017-09-25
Official OpenVPN Connect Android client error with PiVPN on Stretch: Requested encryption or digest alg not available


killermosi:
_Apologies for posting this as an issue here, but the IRC is slow, and we really need a forum or something where the information is more persistent than on IRC anyway._

I have upgraded my Pi OS from Jesse to Stretch a few days ago and finally got around to install PiVPN on the Pi, using the recommended installation method:

`curl -L https://install.pivpn.io | bash`

Everything went smooth, with the exception of a warning screen that the installed OS is not explicitly supported (which was no surprise).

After installation, I have created a new `.ovpn` profile for my Android device, and imported in the the official OpenVPN client: `OpenVPN Connect`.

However, when attempting to connect (after entering the password), this error popped out:

`CORE_ERROR PolarSSL: error parsing config private key : PKCS5 - Requested encryption or digest alg not available`

I did some digging, and the cause is that the encrypted private key in the `.ovpn` file uses a newer format that is not supported by the current Android OpenVPN client.

There are several options:

1. Don't use a password when creating the `.ovpn` file (**NOT** recommended!)
1. Downgrade to Jesse.
1. Wait for an upgrade of the Android OpenVPN client. Should happen sooner than later, but it may take a while.
1. Use an alternative client that supports the newer format. Apparently [OpenVPN for Android](https://play.google.com/store/apps/details?id=de.blinkt.openvpn&hl=en) by Arne Schwabe does so. BEWARE: Untested, use at your ow risk.
1. Re-encode the private key in the `.ovpn` file to an older format that is understood by OpenVPN Connect

For the more impatient people, like me, here is how to re-encode the private key:
```
cd ~/ovpns
cp not-working-profile.ovpn not-working-profile-backup.ovpn
openssl rsa -aes256 -in not-working-profile.ovpn -out new-key-file.key
```

This will ask for the key password and will generate a new key file (asking for a new password - you can use the same one), containing ONLY the key, having the following header:
```
-----BEGIN RSA PRIVATE KEY-----
Proc-Type: 4,ENCRYPTED
DEK-Info: AES-256-CBC,XXXXXXXXXXXXXXXXXXXXXXXXXXXXXX

<private key content follows>
```

To compare, the old private key header (from the `.ovpn` file) looks like this:
```
-----BEGIN ENCRYPTED PRIVATE KEY-----
<private key content follows>
```

[...]

PR #402, opened 2017-11-04
Script for ovpn file conversion for the "OpenVPN Connect" app (iOS+Android)

TimmThaler:
Based on the [issue 364](https://github.com/pivpn/pivpn/issues/364) I created a script that does convert the original ovpn file of a client from an rsa to aes256 key so that it will be usable with both the iOS and Android app "OpenVPN Connect".

I would appreciate some testing. 👍 

Edit: Script needs to be run as sudo

### DIFF
--- a/pivpn
+++ b/pivpn
@@ -69,6 +69,10 @@ function versionFunc {
     printf "\e[1mVersion 1.9\e[0m\n"
 }
 
+function convertIOSFunc {
+    $SUDO /opt/pivpn/convertiOS.sh
+}
+
 function helpFunc {
     echo "::: Control all PiVPN specific functions!"
     echo ":::"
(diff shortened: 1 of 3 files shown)


TimmThaler:
Just updated the script so that all clients are shown now. There was a wrong variable name... I also added an exit point when no client is chosen.

Any testing and feedback is welcome 😃

TimmThaler:
While this is an issue for both the iOS and the Android app "OpenVPN Connect" I changed the script name and the descriptions.

justinamcafee:
Is this pull still needed? I show that OpenVPN connect updated to 3.0 around Feb/Mar 18 and that the conversion to aes256 is no longer necessary.

cfcolaco:
@justinamcafee yup think this pull isn't needed anymore and never agreed with providing such a feature, anyway. 

@redfast00 closing this one, please tell me if anything wrong with it.

PR #412, opened 2017-11-19
Issue-171: Error using iOS OpenVPN

tuaminx:
As @fyellin in the comments, there is some chatter on other groups that some OpenVPN implementations cannot handle client keys that are encrypted with RSA.

If the client key is encrypted, we might re-encrypting the current client key using 3DES.

This commit will convert user client key to 3DES in command `pivpn -a`

P/S: All credits to @fyellin. Many thanks to him.

### DIFF
--- a/scripts/makeOVPN.sh
+++ b/scripts/makeOVPN.sh
@@ -121,6 +121,17 @@ function keyPASS() {
     expect eof
 EOF
 
+    #Convert key to des3
+    KEY_FILE="pki/private/${NAME}${KEY}"
+    expect << EOF
+    set timeout -1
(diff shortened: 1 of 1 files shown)


redfast00:
LGTM. Have you tested this? @cfcolaco: what do you think?

cfcolaco:
@redfast00 TBH till the end of the year I will be probably struggling to keep my nose on top of the water ... so many stuff to do .. barely have time to breathe ... =/

tuaminx:
@redfast00 yes, I use it on my RPi3 before making a PR :)

PR #432, opened 2017-12-19
Updated Issue Template

cfcolaco:
Noticed the lack of information from PiVPN users when opening issues, some of that information is quite useful to troubleshoot, so I have changed the issue template to provide us with more information than the previous one!  and added a few comments in order to to try to raise awareness for users that they need to provide us with accurate information, not just a simple "It doesn't work!"

Thanks


redfast00:
Solid work, thank you!

Issue #426, opened 2017-12-04
Installation fails on ubuntu 14.


username-is-already-taken2:
Hi there

I've hit a bug whilst attempting to install pivpn.

I'm using the `https://app.vagrantup.com/bento/boxes/ubuntu-14.04` as a base image so if you use vagrant you should be able to replicate, here is my vagrantfile

```
Vagrant.configure("2") do |config|
  config.vm.box = "bento/ubuntu-14.04"
  config.vm.provider "vmware_workstation" do |v|
    v.gui = true
  end
  config.vm.box_check_update = false
end
```

I've started the vm, ran a `apt update` & `apt upgrade` and then when I go

::: Your system is up to date! Continuing with PiVPN installation...
Can't call method "set" on an undefined value at /usr/share/perl5/Debconf/FrontEnd.pm line 126,  to call the script this happens

```
root@vagrant:~# curl -L https://install.pivpn.io | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   178  100   178    0     0    302      0 --:--:-- --:--:-- --:--:--   302
100 53505  100 53505    0     0  71569      0 --:--:-- --:--:-- --:--:-- 71569
:::
::: You are root.
::: Verifying free disk space...
:::
::: Checking apt-get for upgraded packages.... done!
:::
::: Your system is up to date! Continuing with PiVPN installation...
Can't call method "set" on an undefined value at /usr/share/perl5/Debconf/FrontEnd.pm line 126, <GEN0> line 5.```

I could try a fresh install using the iso if you think it might be the box, but I use them for a lot of other tasks.

cfcolaco:
Ok .. Before I test this and check this .. why do you want Ubuntu 14?

Any special requirement? 
PiVPN should work on ubuntu 14 ... however... I consider this to be kind of a waste of time since its already quite old release =/

however don't take me wrong... if you really need that ... we can try to work it out!! Just be aware that I don't have much time to help you out.

username-is-already-taken2:
Hi @cfcolaco Cheers for coming back to me

I'm not tied to that version so don't want to waste your time, I think I only choose that because that's what it said on the homepage

`FLEXIBLE ::: ...the installer will work just as well on an Ubuntu Server running Trusty Tahr 14.04` 

Let me quickly stand up a v16 box and try that.

username-is-already-taken2:
OK here is the result from 16.04

[...]

PR #433, opened 2017-12-20
'debconf-apt-progress' seems to be broken on ubuntu

username-is-already-taken2:
PR to resolve issue #426 

stopped using `debconf-apt-progress` on ubuntu as it seems to be broken.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -474,7 +474,13 @@ install_dependent_packages() {
   echo iptables-persistent iptables-persistent/autosave_v6 boolean false | $SUDO debconf-set-selections
 
   if command -v debconf-apt-progress &> /dev/null; then
-    $SUDO debconf-apt-progress -- ${PKG_INSTALL} "${argArray1[@]}"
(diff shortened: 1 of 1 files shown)


cfcolaco:
Thank you a lot @username-is-already-taken2  ... Can this solve more issues we have around here?? 

Seems good to merge into test branch... however didn't have the time to test and replicate this and now that we know that the package is causing issues .. I'd rather see it this as a more permanent solution rather the package and keeping users in the dark regarding installation progress ... IMO this is more "temporary" fix ...

Has the package been ditched? has it changed name? is it now part of another package?
think that replacing it with the correct package name and version would be more appropriate... 

@redfast00 @0-kaladin what you think about it??

cfcolaco:
@username-is-already-taken2 made some more testing and it seems that the issue is a far more specific thing than a general issue ... this being said ... closing the PR. 

Feel free to make a new PR if you notice that something needs fixing from our side!! =) 

Thank you a lot @username-is-already-taken2

PR #455, opened 2018-01-25
Basic changes--a random subnet

adrianQwest:
NB new to github so this is for consideration, probably not quite ready for merging.

I had a problem where the VPN subnet clashed with the wireless network subnet I was using (free wifi). At least that's what I thought the problem was.

I've modified the install script to generate a random subnet rather than the often encountered 10.8.0.0. The sub net is of the form 10.X.Y.0 where X and Y are between 29 and 229 avoiding the common 10.Z.0.0 where Z is usually a small number.

Add generation of random subnet.
This adds a random subnet rather than 10.8.0.0.
It writes the subnet to a file so that it can be restored if necessary.

Let me know what you think. I've used the scipt wto genater new instances of pivpn and it works for me. 

I need to modify the fix iptables routine as the debug thows up a problem. Not sure if I've done the right. Possibly the debug function and fix iptables needs to change.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -58,6 +58,12 @@ IPv4gw=$(ip route get 8.8.8.8 | awk '{print $3}')
 availableInterfaces=$(ip -o link | grep "state UP" | awk '{print $2}' | cut -d':' -f1 | cut -d'@' -f1)
 dhcpcdFile=/etc/dhcpcd.conf
 
+#set address for subnet
+IPDIG1=$(( $RANDOM % 201 + 29))
+IPDIG2=$(( $RANDOM % 201 + 29))
+IPv4SubNet="10.$IPDIG1.$IPDIG2"
+echo $IPv4SubNet > /etc/pivpn/vpnSubNet.txt
+
 # Next see if we are on a tested and supported OS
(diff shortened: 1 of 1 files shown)


cfcolaco:
It looks good to me, and it doesn't interfere with the changes I am doing... @redfast00 it's up to you ...

redfast00:
@adrianQwest: thanks for contributing! Ping me once you feel your code can be merged and you've tested it, and I'll merge it.

cfcolaco:
btw @redfast00 .... just don't merge to master branch ... think that in the meanwhile we should only merge if it is something important or related info provided to users! 

the latest commits to the master branch are already in the test branch as well ...

adrianQwest:
Hi,
I think I have got the debug to work. I'll test then add that as well.

Should I for the test branch and add things there rather than the master? (Just getting to grips with GitHub.)

cfcolaco:
@adrianQwest it doesn't matter for now! you can just use the master and we send the changes to test! =) everything that was changed in the master branch is already pushed into the test!

adrianQwest:
I'll close this pull request and re-submit to the test branch when I have tested the debug script.

PR #471, opened 2018-02-15
2.4 Updates

orazioedoardo:
This pull request contains an option for the user to add openvpn 2.4 features:

- tls-crypt instead of tls-auth to encrypt (not only authenticate) the control channel
- elliptic curves for the key exchange (no need to generate/download dh parameters)
- compress lz4 instead of compl-lzo for less cpu intensive compression

I also changed the logging behavior: current implementation only keeps the log until daemon restart, now openvpn uses rsyslog so we can easily enable log rotation (see confLogging function). 

P.S Sorry for the other wrong pull request

### DIFF
--- a/Default.txt
+++ b/Default.txt
@@ -13,4 +13,4 @@ verify-x509-name SRVRNAME name
 cipher AES-256-CBC
 auth SHA256
 comp-lzo
-verb 1
+verb 3
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -852,25 +852,36 @@ EOF
     # Build the server
     ${SUDOE} ./easyrsa build-server-full ${SERVER_NAME} nopass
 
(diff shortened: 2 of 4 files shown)


redfast00:
Thank you for contributing. Do all major OpenVPN apps work with these changes (so at least: `OpenVPN for Android`, and `OpenVPN Connect` on Android and iOS)?

redfast00:
Thanks for contributing, looks good to me. This seems ready to merge, have you tested this?

orazioedoardo:
@redfast00 
I’ve tested with mobile clients and a debian stretch vm.

@d0nlab 
Clone the repository: `git clone -b test https://github.com/orazioedoardo/pivpn.git`
Change [this line](https://github.com/pivpn/pivpn/blob/test/auto_install/install.sh#L49) from false to true.
Run the install script.
Tell us if it works for you as well.

Issue #473, opened 2018-02-22
comp-lzo is no longer valid


lookfirst:
<img width="617" alt="tunnelblick 2018-02-22 10-46-19" src="https://user-images.githubusercontent.com/85355/36519414-f1a29026-17bd-11e8-86fd-c00f5e0f29ad.png">

cfcolaco:
This is solved with pr #471 

Discussion regarding this being tracked at #473

teake:
Sorry to reopen this, but I I just re-installed PiVPN on my Raspberry Pi 3 and am still getting the above warning. 

I double-checked the discussion in #474 and realized that I didn't get the `Version 2.4 improvements` whiptail dialog during install. There's a conditional around that dialog that prevents it from showing in Raspbian: 

https://github.com/pivpn/pivpn/blob/33e7ef0c9191c86f74d6b7bc06371350d49e3af9/auto_install/install.sh#L839 

What's the recommended way to get 2.4 improvements in Raspbian?

teake:
Ok, nevermind, I'm on Raspbian Jessie and my OpenVPN version is 2.3.4. Looks like I'll be updating to Raspbian Stretch (see also #280).

PR #474, opened 2018-02-22
comp-lzo is deprecated. #473

lookfirst:


cfcolaco:
Thanks a lot @lookfirst , but you should have searched for that first, pr #471 Fixes this. being already effective in the Main PiVPN Branch. 

Have you tested this? Have you noticed the newer changes on the main branch? 
More information in the description would be more helpful.  

@orazioedoardo Is the lz4 optional in your pr? Perhaps we should make it default ... 

if you have the time, can you add some info regarding this in the README please?? 

Or perhaps send it to me and I can put it in the wiki part of the repo.

Unfortunately, I didn't have the opportunity to test your PR and new features. =/

orazioedoardo:
We could indeed make LZ4 default but that implies openvpn 2.4. I’d wait for full ECDSA support among all platforms though.
At that point we can simply move [this](https://github.com/pivpn/pivpn/blob/master/auto_install/install.sh#L856) dialog at the top of the `confOpenVPN` function such that we branch like this:

“Last features” (recommended)

- ecdsa certificates
- ecdh key exchange
- tls-crypt
- lz4 compression

or “Compatibility”

- rsa cerificates
- dh key exchange
- tls-auth
- lzo compression

lookfirst:
@cfcolaco #471 didn't make this change. See: https://github.com/pivpn/pivpn/pull/471/files#diff-ce09a633c44342936b9e6c2c47b890caR15

Here is what happened to me. I have automatic beta updates turned on for Tunnelblick. After coming back from a nice vacation in Bali, I upgrade my desktop to `Tunnelblick 3.7.5beta08 (build 5003)` and connect to my two vpn's (both running this wonderful project) and the connections are not stable at all. They only stay up for a couple minutes at a time and then drop out. Nothing shows up in the logs that seems to signify why. I also get the screenshot described in #473 .

So, I 'upgrade' the first vpn, load the new ovpn file and I see the same error again. I notice Default.txt is still generating the old config so I come here, do a bit of searching, notice #471 has been merged and file the issue with the screenshot to let you know that client software is affected as well as try to be helpful with a simple one line PR.

Note, I'm a novice around OpenVPN and I honestly don't want to become an expert. That's why I use your project to begin with! :-)

Needless to say, after upgrading both VPN (which actually took a bit of work cause I didn't realize uninstalling would wipe my nicely setup `server.conf` file... grrr... ) and hacking the `Default.txt` file, things are stable again.

On a separate note (feel free to create a new issue if you care enough), on my regular Ubuntu stretch/sid box, I had to manually add the apt repo for OpenVPN in order to get the latest version (2.4.4) and on my actual Pi vpn running raspbian 9.1, I can't seem to get better than 2.4.0. It would be nice if your scripts found the latest versions for me.

orazioedoardo:
@lookfirst Can you test this? Default client and server configs are the same but the script should change compression algorithm [here](https://github.com/pivpn/pivpn/blob/master/auto_install/install.sh#L899) and [here](https://github.com/pivpn/pivpn/blob/master/auto_install/install.sh#L1014) if you answer “Yes” [here](https://github.com/pivpn/pivpn/blob/master/auto_install/install.sh#L856)

eccgecko:
I think default compress option should be lz4-v2 instead of lz4. The difference is pretty small, but v2 apparently has slightly better packet data alignment than plain lz4, and every little helps on a pi. See here [https://community.openvpn.net/openvpn/ticket/820](url)
I have changed all instances of lz4 in my config files to lz4-v2 and have seen no issues arise as a result of this change, so it shouldn’t be a problem to make this default in the script.

PR #493, opened 2018-03-09
Updated install.sh

justinamcafee:
Added support for FamilyShield DNS from OpenDNS/CISCO. 
With a very minor code change, individuals can now implement restrictions on dangers, disturbing, or otherwise adult oriented content without the need for managed restrictions. 
This is a fairly non-invasive change and will benefit users who intend to use VPN for home or small business uses where access to such material may be undesirable.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -695,6 +695,7 @@ setClientDNS() {
             Level3 "" off
             DNS.WATCH "" off
             Norton "" off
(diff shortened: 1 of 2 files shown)


justinamcafee:
I have ran the code and verified that the additional servers are added to the output of /etc/openvpn/server.conf when selected.

redfast00:
Thank you!

PR #359, opened 2017-09-22
Removed the duplicate-cn option so every device/connection must use a unique certificate

jellemdekker:
Removed the duplicate-cn option so every device/connection must use a unique certificate. Please see issue #358 for an in-depth explanation.

PR #494, opened 2018-03-09
Added Optional Duplicate Certificate screen in installer. 

justinamcafee:
Often users in a home or business environment may desire to issue VPN certificates on a per-user basis. 
Adding duplicate-cn as an option allows the users to choose the best set-up for them. As the "client-config-dir" option is not currently supported by default, adding this option only empowers our users, and does not detract from their abilities or interfere with other functions. 

Use Cases include: 
* Home where parents may wish to provide one certificate per family member who may have multiple devices and revoke or disable as necessary. (Johnny has a laptop and cellphone. Johnny is grounded, revoke Johnny's certificate.) 
* Business where admins may wish to provide one certificate per user with multiple devices and revoke or disable as necessary. (John has a laptop and cellphone. John is fired, revoke John's certificate.)

While this may be seen as a less desirable option, the default and tagged [recommended] option is for duplicate-cn to be commented out of server-conf.txt.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -798,7 +798,7 @@ setOptDuplicate() {
     DupeCNCmd=(whiptail --separate-output --radiolist "Will certificates be distributed on a per-client or per-device[recommended] basis?" ${r} ${c} 6)
     DupeCNChooseOptions=(Per-Device "" on
             Per-Client "" off)
-    if DupeCNchoices=$("${DupeCNCmd[@]}" "${DCNChooseOptions[@]}" 2>&1 >/dev/tty)
+    if DupeCNchoices=$("${DupeCNCmd[@]}" "${DupeCNChooseOptions[@]}" 2>&1 >/dev/tty)
     then
         case ${DupeCNchoices} in
 	Per-Device)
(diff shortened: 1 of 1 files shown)


justinamcafee:
@redfast00  @0-kaladin  Didn't realize that your PR's are directed towards the test branch, per Contributing.md. Want me to put this there? It works fine on anything that passes the supported OS requirements.

redfast00:
I don't really think this is in scope of the PiVPN project. Any thoughts @cfcolaco?

justinamcafee:
@cfcolaco @redfast00  This was originally a part of the project. The addition I've made only replaces the option to explain the configuration options to the user during install. This is in response to #359 which forced non-duplicate certificates without cause.

cfcolaco:
@justinamcafee part of the original project or not, it was removed. 

I can recognize that use for such an option, however, 

I believe that the use cases for such an option the minority of who uses this script. 

In My point of view, this script helps out people with less knowledge to have a VPN server, and allowing duplicate certificates to be used, is a big security // privacy concern. 

Empowering users to do so seems a bit unreasonable. 

In my opinion, whoever wants to have that option enabled will know exactly why and how to enable it. 

OBS: If we are going to provide such an option we should as well provide an easy way to revert that back in case someone enables that by mistake. 

would be nice to know what @0-kaladin thinks about this. 
@redfast00 its up yo you ... as I say I don't entirely agree with this.

redfast00:
Yeah, that's about what I thought. Will close this then.

PR #498, opened 2018-03-14
AES Encryption instead of 3DES

Zeik0s:
I was wondering why in the script the encryption method 3DES is used. This algorithm is long outdated and deprecated. By simply choosing the AES with 128 bits key length the saved key would be safer for being brute-forced. Therefore it would be a security bonus without changing much.


InnovativeInventor:
This sounds like a no-brainer, I like this!

redfast00:
First off, thanks for contributing! I think we were using 3DES for compatibility with iPhones or something. A quick seach through the open issues found https://github.com/pivpn/pivpn/issues/171.

orazioedoardo:
@redfast00 AES128 works fine with iPhones (AES256 works too).

redfast00:
@orazioedoardo thanks. I can now merge this, but this PR contains a lot of unrelated commits and I don't think it's been tested yet. Any input @cfcolaco ?

cfcolaco:
@redfast00 think this PR is actually cool ... will merge into test all commits ahead of Test branch! so should be fairly safe to commit! =)

trilleplay:
so are we merging this, i'm all for it

PR #501, opened 2018-03-16
DNS leak fix in server_config.txt

Cyb0rk:
Added fix to prevent DNS leak on Windows systems. If the OS on the client side is Linux, this option is ignored by the client.

### DIFF
--- a/server_config.txt
+++ b/server_config.txt
@@ -10,6 +10,8 @@ server 10.8.0.0 255.255.255.0
 # Set your primary domain name server address for clients
 push "dhcp-option DNS 8.8.8.8"
 push "dhcp-option DNS 8.8.4.4"
+# Prevent DNS leaks on Windows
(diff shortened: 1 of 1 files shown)


cfcolaco:
Seems good to me. 
Tough some care with this option might be required. 

> --block-outside-dns
> Block DNS servers on other network adapters to prevent DNS leaks. This option prevents any application from accessing TCP or UDP port 53 except one inside the tunnel. It uses Windows Filtering Platform (WFP) and works on Windows Vista or later.
> This option is considered unknown on non-Windows platforms and unsupported on Windows XP, resulting in fatal error. You may want to use --setenv opt or --ignore-unknown-option (not suitable for Windows XP) to ignore said error. Note that pushing unknown options from server does not trigger fatal errors.

redfast00:
Looks good to me, thanks @Cyb0rk!

PR #514, opened 2018-04-02
Several changes and some fixes

orazioedoardo:
- Simplified [distro_check](https://github.com/orazioedoardo/pivpn/blob/master/auto_install/install.sh#L82) function.
- New [addSoftwareRepo](https://github.com/orazioedoardo/pivpn/blob/master/auto_install/install.sh#L407) function to add the openvpn repo. Actually [checkForDependencies](https://github.com/orazioedoardo/pivpn/blob/master/auto_install/install.sh#L505) was supposed to do this but it is [commented out](https://github.com/orazioedoardo/pivpn/blob/master/auto_install/install.sh#L1284), looks like it had been spit into individual functions later.
- Simplified [DNS server](https://github.com/orazioedoardo/pivpn/blob/master/auto_install/install.sh#L709) choise.
- Do not prompt to enable 2.4 features if Raspbian [is not](https://github.com/orazioedoardo/pivpn/blob/master/auto_install/install.sh#L839) stretch.
- Explicitly state the [curve](https://github.com/orazioedoardo/pivpn/blob/master/auto_install/install.sh#L892) to use for ECDH (should not be required but it makes more clear what's going on).
- If ufw is installed but not enabled and the system language is not [english](https://github.com/orazioedoardo/pivpn/blob/master/auto_install/install.sh#L947) the installer would detect that the firewall is enabled.
- Changed [confLogging](https://github.com/orazioedoardo/pivpn/blob/master/auto_install/install.sh#L1058) to make github's syntax highlighting happy.
- Fixed permission denied on [finalExports](https://github.com/orazioedoardo/pivpn/blob/master/auto_install/install.sh#L1087).
- Adjusted indentation.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -82,47 +82,33 @@ Would you like to continue anyway?" ${r} ${c}) then
 distro_check() {
   # if lsb_release command is on their system
   if hash lsb_release 2>/dev/null; then
-      PLAT=$(lsb_release -si)
-      OSCN=$(lsb_release -sc) # We want this to be trusty xenial or jessie
 
-      case ${PLAT} in
-          Ubuntu|Raspbian|Debian|Devuan)
-          case ${OSCN} in
-              trusty|xenial|jessie|stretch)
-                  ;;
-              *)
-                  maybeOS_Support
-                  ;;
-          esac
+    PLAT=$(lsb_release -si)
+    OSCN=$(lsb_release -sc) # We want this to be trusty xenial or jessie
+
+  else # else get info from os-release
+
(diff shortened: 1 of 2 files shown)


cfcolaco:
How much of it have you tested?
Any issues get potentially fixed with this? 

Apart from that, to me it seems a really great job! =)

orazioedoardo:
This pull request was supposed to contain a fix for missing repositories and the check for the Raspbian version (many people with raspbian jessie complained about this, they enabled 2.4 improvements but their openvpn version was not supported), however I added more stuff in the process.
I’ve tested the installation on Raspbian Stretch, Debian Stretch and Ubuntu 16.04. It shouldn’t break anything but would be nice if someone tested this as well.

redfast00:
Thanks!

PR #538, opened 2018-05-19
Add undocumented avoid_static_ip flag

### LINKED ISSUE #536: Make static IP setup optional
Hi, 

thanks for making this awesome project! It works really great on my Raspberry Pi 3. I only have one little feature request:

Would it be possible to make the setup of the static IP during the installation option? 

Something like this must be easily possible if I see this function: `avoidStaticIPv4Ubuntu`

I may make a pull request myself if I find the time.

Thank you and continue your awesome work.

Best, 
Chris

cpoetter:
Hi,

As promised my pull request to close #536.

Best, 
Chris

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -48,6 +48,7 @@ c=$(( c < 70 ? 70 : c ))
 skipSpaceCheck=false
 reconfigure=false
 runUnattended=false
+skipStaticIpSetup=false
 
 # Find IP used to route to outside world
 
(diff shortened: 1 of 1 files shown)


Zeik0s:
Well, the question is, why would anyone do that? You would have to change the IP in your config file everytime, instead of simply connecting to your server. If you have a public ip from your ISP, you can still connect via dynamic DNS services.

cpoetter:
@Zeik0s Well, I think my use case is not very common, but maybe some others need it for similar things, so I thought an undocumented flag would be a good solution for this.

**My use case:**
I don't care about the internal network the Pi is connected to. I only care about the external (American) IP the internal network is routed through. Therefore I would like the Pi to be flexible enough to react to changing local networks/IP address ranges. The clients of the VPN can find the VPN server via a reverse SSH tunnel also running on the Pi. And this IP is the actual static IP for the clients.

Best, 
Chris

redfast00:
I don't think this is within the scope of PiVPN: this is for advanced users/a really specific usecase. If users want to uses this, they should just edit the files manually.

PR #572, opened 2018-08-03
Adds Cloudflare's DNS Privacy First option

juanse417:
I thought it might be helpful to suggest also Cloudflare's DNS option. As they promote a Privacy DNS service. I personally use them in my my setup.


redfast00:
Thank you, this has been done by another user in #583

Issue #588, opened 2018-08-23
No public ip address visible during install


corkerss:
# PiVPN Issue Template

During installation no public ip address is visible. Only a blank space

## Have you searched for similar issues and solutions? 
  yes, public ip address issues
<img width="649" alt="screen shot 2018-08-23 at 22 13 46" src="https://user-images.githubusercontent.com/36156736/44552614-bf98f800-a722-11e8-9ab4-6861a21a1feb.png">

  

## Console output of `curl -L install.pivpn.io | bash`
```
 % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   178  100   178    0     0    572      0 --:--:-- --:--:-- --:--:--   572
100 51064  100 51064    0     0  91568      0 --:--:-- --:--:-- --:--:-- 91568
:::
::: sudo will be used for the install.
```

## Console output of `pivpn add` or `pivpn add nopass`
```
  Output Here
```

## Console output of `pivpn debug`
```
::: Generating Debug Output
:::					:::
::		PiVPN Debug		 ::
:::					:::
::	Latest Commit			 ::
:::					:::
commit de76940b65bab86eda772158da58328482363559
Merge: 18bfa92 58fcaaf
Author: redfast00 <redfast00@users.noreply.github.com>
Date:   Wed Aug 15 21:17:33 2018 +0200

    Merge pull request #554 from Giraffe1966/issue-552
    
    Fix issue #552 (remove dead code)
:::					:::
::	Recursive list of files in	 ::
::	/etc/openvpn/easy-rsa/pki	 ::
:::					:::
/etc/openvpn/easy-rsa/pki/:
ca.crt
crl.pem
Default.txt
index.txt
index.txt.attr
index.txt.old
issued
private
serial
serial.old
ta.key

/etc/openvpn/easy-rsa/pki/issued:
server_nD8T35MxdmUHaNSg.crt

/etc/openvpn/easy-rsa/pki/private:
ca.key
server_nD8T35MxdmUHaNSg.key
:::					:::
::	Output of /etc/pivpn/*		 ::
:::					:::
:: START /etc/pivpn/DET_PLATFORM ::
Debian
:: END /etc/pivpn/DET_PLATFORM ::
:: START /etc/pivpn/INSTALL_PORT ::
11942
:: END /etc/pivpn/INSTALL_PORT ::
:: START /etc/pivpn/INSTALL_PROTO ::
udp
[...]

Issue #567, opened 2018-07-27
CRL has expired


svib:
After having worked splendidly for about half a year my piVPN server installation refuses connecting clients.  From the server log file, it seems that the CA certificate has expired. I have looked around and there are some guides how to renew it, but nothing specific for the piVPN installation, so it is not directly clear how to apply it.  Maybe this could be an option in the pivpn menu? Any suggestion how to resolve it?

```
Fri Jul 27 13:07:46 2018 us=790264 MULTI: multi_create_instance called
Fri Jul 27 13:07:46 2018 us=790712 192.168.4.100:53808 Re-using SSL/TLS context
Fri Jul 27 13:07:46 2018 us=790800 192.168.4.100:53808 LZO compression initializing
Fri Jul 27 13:07:46 2018 us=791129 192.168.4.100:53808 Control Channel MTU parms [ L:1622 D:1172 EF:78 EB:0 ET:0 EL:3 ]
Fri Jul 27 13:07:46 2018 us=791215 192.168.4.100:53808 Data Channel MTU parms [ L:1622 D:1450 EF:122 EB:406 ET:0 EL:3 ]
Fri Jul 27 13:07:46 2018 us=791412 192.168.4.100:53808 Local Options String (VER=V4): 'V4,dev-type tun,link-mtu 1570,tun-mtu 1500,proto UDPv4,comp-lzo,keydir $
Fri Jul 27 13:07:46 2018 us=791483 192.168.4.100:53808 Expected Remote Options String (VER=V4): 'V4,dev-type tun,link-mtu 1570,tun-mtu 1500,proto UDPv4,comp-l$
RFri Jul 27 13:07:46 2018 us=791670 192.168.4.100:53808 TLS: Initial packet from [AF_INET]192.168.4.100:53808, sid=e68cd7e3 4f098689
WRRWWRRWRFri Jul 27 13:07:46 2018 us=932679 192.168.4.100:53808 VERIFY ERROR: depth=0, error=CRL has expired: CN=XXXXXXX
Fri Jul 27 13:07:46 2018 us=932886 192.168.4.100:53808 OpenSSL: error:14089086:SSL routines:ssl3_get_client_certificate:certificate verify failed
Fri Jul 27 13:07:46 2018 us=932924 192.168.4.100:53808 TLS_ERROR: BIO read tls_read_plaintext error
Fri Jul 27 13:07:46 2018 us=932958 192.168.4.100:53808 TLS Error: TLS object -> incoming plaintext read error
[...]

Issue #526, opened 2018-04-30
DNS on host disabled(?) after pivpn installed


mnbf9rca:
# PiVPN Issue Template
``
## Issue
After installation, DNS stops working on the host (DNS configuration is performed by dhcpd at boot time). This is a clean install on ubuntu 18.04LTS on linode. The only configuration which was performed after OS installation is to add a user to store the pivpn data (adduser robert) and that user was added to sudo group (usermod -aG sudo robert). To be clear, DNS was working (via dhcpd) before installation.

After reboot i see behaviour similar to the following:
```
root@localhost:~# nslookup www.google.com
;; connection timed out; no servers could be reached

root@localhost:~# ping 8.8.8.8       
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=47 time=7.77 ms
64 bytes from 8.8.8.8: icmp_seq=2 ttl=47 time=7.89 ms
64 bytes from 8.8.8.8: icmp_seq=3 ttl=47 time=7.82 ms
^C
--- 8.8.8.8 ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2003ms
rtt min/avg/max/mdev = 7.771/7.830/7.892/0.049 ms
root@localhost:~# nslookup 
> server 8.8.8.8
Default server: 8.8.8.8
Address: 8.8.8.8#53
> www.google.co.uk
Server:         8.8.8.8
Address:        8.8.8.8#53

Non-authoritative answer:
Name:   www.google.co.uk
Address: 216.58.201.3
Name:   www.google.co.uk
Address: 2a00:1450:4009:805::2003
```

## Have you searched for similar issues and solutions? 
    (yes/no / which issues?) 
  Yes. https://github.com/pivpn/pivpn/issues/506 but this looks like it's not quite the same (here DNS is missing for the clients)

## Console output of `curl install.pivpn.io | bash`
```
root@localhost:~# curl -L https://install.pivpn.io | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   178  100   178    0     0    265      0 --:--:-- --:--:-- --:--:--   265
100 55909  100 55909    0     0  69799      0 --:--:-- --:--:-- --:--:-- 69799
:::
::: You are root.
::: Did not detect perfectly supported OS but,
::: Continuing installation at user's own risk...
[...]

Issue #181, opened 2017-01-08
Why was easy-rsa forked?


dp87:
I was looking at the [install.sh](https://github.com/pivpn/pivpn/blob/master/auto_install/install.sh) script, and I was wondering if it was really necessary to use [pivpn's own easy-rsa fork](https://github.com/pivpn/pivpn) rather than the [official one](https://github.com/OpenVPN/easy-rsa/tree/master/easyrsa3)?

What was the reason behind this choice?

redfast00:
We want to eventually use ECDSA. Since the easy-rsa package in Debian is outdated, we decided to host it ourselves.

dp87:
Understood, makes sense.

devZer0:
i don't understand. what has "outdated in debian" to do with forking / bringing own version ? why not using https://github.com/OpenVPN/easy-rsa instead ?

it looks ecdsa is in there for years, at least according to https://github.com/OpenVPN/easy-rsa/commit/cc65460437554409dee7592ca36d3ed6f1db057f

Giraffe1966:
@devZer0 The official easy-rsa package is now used in the script. See #597.

devZer0:
what about deleting the forked github repo then ?

PR #597, opened 2018-09-04
ECDSA support plus several fixes

orazioedoardo:
1. The openvpn 2.4 dialog has been changed. Now if the user chooses “Yes” the installer will use ECDSA (256, 384, 521 bit with 256 being the default), ECDH (openvpn will use the curve from the certificate) and tls-crypt. Otherwise the usual RSA (1024, 2048, 4096), matching DH parameters and tls-auth.
2. Package dhcpcd5 will be installed only if the distribution is not Ubuntu Bionic (see #526). Actually i haven’t tested ubuntu 18.04 so I Ieft its status at “maybeos_support”.
3. Replaced the easy-rsa download with the official (and latest) one.
4. Simplified distro_check.
5. Linux human users start at UID 1000 so I changed the chooseuser function accordingly, otherwhise invalid users like “pihole” would show up.
6. Packages installation would fail with no visible error if the distribution uses https repositories (like dietpi), so I fixed install_dependent_packages.
7. Unattended upgrades were ineffective on raspbian since there is not “Raspbian-Security” label. I noticed this few weeks ago because a critical openssh-server update (regarding username enumeration) had not been installed. I changed the script to download the aprropriate config file ([more info](https://raspberrypi.stackexchange.com/questions/72022/configuring-unattended-upgrades-on-raspbian-stretch)).
8. Set expiry date of the CRL to 10 years (see #567).
9. According to issue #588 dig may not return the public IP so I changed the script to check the output falling back to curl if it was empty or invalid. eth0.me is also not available so I changed it to checkip.amazonaws.com. Should also fix a common issue that makes the script skip to the end leaving a broken installation (see https://github.com/pivpn/pivpn/issues/617).
10. Removed compression as it could be a vector of attacks (see [this](https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg16919.html)).
11. Workaround that re-encrypts the private key is not required anymore as OpenVPN Connect seems to work with the default openssl encrpytion.

Opinions are welcome

### DIFF
--- a/Default.txt
+++ b/Default.txt
@@ -12,5 +12,4 @@ tls-version-min 1.2
 verify-x509-name SRVRNAME name
 cipher AES-256-CBC
 auth SHA256
-comp-lzo
 verb 3
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -949,13 +949,11 @@ confNetwork() {
 }
 
 confOVPN() {
-    IPv4pub=$(dig +short myip.opendns.com @208.67.222.222)
-    if [ $? -ne 0 ] || [ -z "$IPv4pub" ]; then
-        echo "dig failed, now trying to curl whatismyip.akamai.com"
-        if ! IPv4pub=$(curl -s http://whatismyip.akamai.com)
-        then
-            echo "whatismyip.akamai.com failed, please check your internet connection/DNS"
-            exit $?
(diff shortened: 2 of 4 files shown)


Giraffe1966:
Why was a fork of easy-rsa created in the first place?

orazioedoardo:
Read #181

PR #652, opened 2018-12-17
fix 'i_do_not_follow_recommendations' behavior + workaround for 'unattended'

pascal-vs:
The commits here fix the use of the flag 'i_do_not_follow_recommendations' and propose a workaround for the flag 'unattended'.

The flag 'i_do_not_follow_recommendations' was not doing what it should do: put 'true' in 'skipSpaceCheck' :) .

The flag 'unattended' on a non tested distribution was still printing a whiptail dialog. I only propose a workaround here: if the flag 'i_do_not_follow_recommendations' is used, 'distro_check' function is avoided. A better fix should print only "echo" warnings or errors in the 'distro_check' function.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1220,7 +1220,7 @@ main() {
     for var in "$@"; do
         case "$var" in
             "--reconfigure"  ) reconfigure=true;;
-            "--i_do_not_follow_recommendations"   ) skipSpaceCheck=false;;
+            "--i_do_not_follow_recommendations"   ) skipSpaceCheck=true;;
             "--unattended"     ) runUnattended=true;;
         esac
     done
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -949,13 +949,11 @@ confNetwork() {
 }
 
 confOVPN() {
-    IPv4pub=$(dig +short myip.opendns.com @208.67.222.222)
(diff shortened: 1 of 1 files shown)


auto_install/install.sh
```diff
@@ -63,13 +64,21 @@ dhcpcdFile=/etc/dhcpcd.conf
...
+    fi
     whiptail --msgbox --backtitle "INVALID OS DETECTED" --title "Invalid OS" "We have not been able to detect a supported OS.
 Currently this installer supports Raspbian and Debian (Jessie and Stretch), Devuan (Jessie) and Ubuntu from 14.04 (trusty) to 17.04 (zesty).
 If you think you received this message in error, you can post an issue on the GitHub at https://github.com/pivpn/pivpn/issues." ${r} ${c}
     exit 1
 }
 
 function maybeOS_Support() {
+    if [["${quietDistroCheck}" = true]]; then
+        echo "::: Exiting due to unsupported OS"
```
4s3ti: Here I think we should better clarify users that the OS Is possibly supported, 
however, for the sake of unattended install,  the script is bailing out because we cannot be sure of it.

4s3ti:
Long time without any action, closing this PR. 
Please make the requested changes and make a new PR. 

Kind regards.

Issue #54, opened 2016-07-21
Trouble connecting through Android OpenVPN Connect on mobile data


AngusF22:
I've had considerable trouble connecting to my Pi VPN while on mobile data. I can ping my phone from the local network when connected, but couldn't do anything else unless the phone was connected to the VPN via Wi-Fi.

I was able to solve this issue after considerable hassle and research by adding two lines to /etc/openvpn/server.conf:

```
tun-mtu 1400 
mssfix 1360
```

With these, I am able to connect the VPN over mobile data, and have complete access to my home network, as well as being able to browse outside the network.

I don't know enough about OpenVPN to say that these are a good idea to add by default. Perhaps someone else who is better versed can weigh in.

0-kaladin:
@AngusF22 I can connect fine over mobile data without those changes.  They seem maybe specific for your mobile network.  In the interest of helping others maybe with the same issue can you respond with your moble provide and device type you used?  Thanks!

AngusF22:
I'm on Rogers, using an LG G3. I forget the exact bit of Googling I did to find those strings, but someone else was having similar issues, and they posted them as a working fix.

0-kaladin:
I'll add a link to this in the wiki so others may find as well.  Thanks!

webdude21:
@AngusF22 After an entire day of fiddling and wondering why internet over the vpn doesn't work, i've stumbled upon your post. Thank you, you made my day!

CheezusChrust:
Thank you for this - this fixed my issue completely. Could somebody smarter than me explain what those commands are doing to actually fix this? It's odd because my mother's phone on the same mobile network has had no issues, however with mine it couldn't stay connected without these settings.

PR #655, opened 2018-12-24
Some enhanced (rights & telekom hybrid)

LMS235:
right adjustment
- read for "other" withdrawn from .ovpn files
- delete also /etc/openvpn/easy-rsa/pki/${CERTS_TO_REVOKE[ii]}.ovpn after remove certificate

Telekom Hybrid Check 
- Small check whether Telekom Hybrid will be used to give helpful tips.

### DIFF
--- a/scripts/makeOVPN.sh
+++ b/scripts/makeOVPN.sh
@@ -241,6 +241,8 @@ echo "tls-auth Private Key found: $TA"
 # Copy the .ovpn profile to the home directory for convenient remote access
 cp "/etc/openvpn/easy-rsa/pki/$NAME$FILEEXT" "/home/$INSTALL_USER/ovpns/$NAME$FILEEXT"
 chown "$INSTALL_USER" "/home/$INSTALL_USER/ovpns/$NAME$FILEEXT"
+chmod o-r "/etc/openvpn/easy-rsa/pki/$NAME$FILEEXT"
+chmod o-r "/home/$INSTALL_USER/ovpns/$NAME$FILEEXT"
 printf "\n\n"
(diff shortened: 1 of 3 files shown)


Giraffe1966:
This seems related to issue #54 (since Telekom Hybrid uses 4G LTE when DSL is congested), which doesn't just affect Telekom Hybrid. I don't believe checking for this specific router is worth it, because it isn't a Telekom-specific problem, and most users won't be affected. Perhaps an informational dialogue similar to the one you've created could be added to the installer warning about cellular usage in general.

LMS235:
Telekom Hybrid uses both connections simultaneously as a tunnel, hence the smaller MTU value. For normal users in DE there is this as far as I know only from the Telekom so and also only these two routers. Professional users should know what to do anyway.

And the check is only visible if one of the two routers is found. So for 98% of the users it is never visible. But for the other 2% surely helpful.

Giraffe1966:
Maybe at the beginning of the installer a `whiptail` prompt could inform people where to search if they have a problem such as this one. It could direct people to check the wiki pages and then search the issues. This could help everybody that has a problem, including Telekom Hybrid users.

LMS235:
This can certainly be installed in addition. But there is nothing against accepting the current pull-request. Or do I understand that wrong?

Giraffe1966:
I think the info prompt should be for all issues, and `chmod 600` would be more secure for the config files (only allowing owner to read/write). Thanks for catching that permission issue!

Giraffe1966:
@LMS235 Please change the target branch to the test branch.

PR #685, opened 2019-03-11
Adding support for custom search domain

quwip10:
Main changes:

  This code creates a new setCustomDomain() procedure to allow users the option to add a custom search domain during install.

It also adds a function valid_domain() that uses regex to do a mild validation on domain input.

This code will then insert into the created server.conf

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -380,6 +380,18 @@ function valid_ip()
     return $stat
 }
 
+#Call this function to use a regex to check user input for a valid custom domain
+function valid_domain()
+{
+  local domain=$1
+  local stat=1
+
+  if [[ $domain =~ ^[a-zA-Z0-9][a-zA-Z0-9-]{1,61}\.[a-zA-Z]{2,}$ ]]; then
+    stat=$?
+  fi
+  return $stat
+}
+
 installScripts() {
     # Install the scripts from /etc/.pivpn to their various locations
     $SUDO echo ":::"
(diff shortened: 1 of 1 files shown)


Giraffe1966:
@quwip10 pivpn is designed to make it as simple as possible to set up a VPN. I'm afraid adding a custom search domain might make things more confusing for inexperienced users. However, only one line has to be changed to add a custom search domain, so it is really easy to add one yourself if you need one.

Giraffe1966:
@quwip10 Please change the target branch to the test branch, so it can be tested first.

quwip10:
I think PiVPN can be useful for all sorts of users rather than just beginners. The custom search domain can often be used with the custom DNS option which is already supported. I also hid the feature behind a warning that it is for advanced users. Pull request added to the test fork although that fork is pretty far behind master.

Giraffe1966:
I've submitted a PR to update the test branch.

Giraffe1966:
@quwip10 I like the idea of adding more advanced features, but only if they are hidden behind advanced feature warnings, like you've suggested. Perhaps pivpn should have an advanced features mode, where more configuration is possible.

quwip10:
@Giraffe1966 I agree. I like the idea of an advanced mode also. I considered coding it, but wanted to get this update in first.

Giraffe1966:
@quwip10 I think this should be closed now, as #706, the PR I believe you intended to merge, has been merged.

PR #705, opened 2019-03-30
Update easy-rsa version

Giraffe1966:
Update easy-rsa version to v3.0.6, which would allow for us to implement a certificate renewal option, since easy-rsa now supports certificate renewal (https://github.com/OpenVPN/easy-rsa/pull/286)!


Giraffe1966:
@orazioedoardo Thanks for spotting this. It has been fixed. Unfortunately, this commit made its way into master.

PR #716, opened 2019-04-10
Fix Easy-RSA install

jamfrade:
Renamed from EasyRSA-{version}.tgz to EasyRSA-unix-v{version}.tgz


james-lasersoft:
Why on earth they can't make up their minds and use the same name inside and outside of the compression is well beyond my comprehension. Thanks for the correction.

Issue #576, opened 2018-08-06
Failing to connect on ubuntu


epikgamerwmpp:
I can not connect on ubuntu and the connection times out.

soydepr:
Show log and reads

https://github.com/pivpn/pivpn/issues/567

devZer0:
ubuntu network manager does not yet support tls-crypt, see https://bugs.launchpad.net/ubuntu/+source/network-manager/+bug/1749562?comments=all  and complain there. they need to fix their network manager

devZer0:
you did not tell which ubuntu version. apparently network manager is already fixed in ubuntu 18.04 - but the problem remains in kubuntu 18.04 as kubuntu is using it's own kde based network manager which suffers from the (former) same problem of not being able to import tls-crypt zert from .ovpn file

PR #728, opened 2019-04-30
Always set tls-crypt instead of tls-auth

alexhenrie:
The install script always installs OpenVPN 2.4, even if it has to
install it from a third-party repository. In 2.4, tls-auth has been
replaced by tls-crypt and tls-crypt must be specified even if using
traditional RSA cryptography.

In short, RSA is currently broken and this patch is necessary to make it work again.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -841,11 +841,6 @@ EOF
     $SUDO cp /etc/.pivpn/server_config.txt /etc/openvpn/server.conf
 
     if [[ ${APPLY_TWO_POINT_FOUR} == true ]]; then
-      #If they enabled 2.4 use tls-crypt instead of tls-auth to encrypt control channel
-      $SUDO sed -i "s/tls-auth \/etc\/openvpn\/easy-rsa\/pki\/ta.key 0/tls-crypt \/etc\/openvpn\/easy-rsa\/pki\/ta.key/" /etc/openvpn/server.conf
-    fi
-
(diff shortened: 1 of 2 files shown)


Giraffe1966:
@alexhenrie I'm pretty sure the option to use tls-auth is to allow older (pre-2.4) clients to still function properly. Also, RSA's and OpenVPN's security is not affected by tls-auth or tls-crypt. tls-auth is mainly useful in preventing DoS attacks by signing every control channel packet with an HMAC signature, while tls-crypt is like tls-auth, but the OpenVPN control channel is encrypted using a key that all peers know, which prevents anybody without the key from viewing the TLS-secured control channel. tls-crypt has all of the benefits of tls-auth, and hides the certificates used in the TLS control channel by encrypting the control channel, and makes the traffic look less like OpenVPN traffic, but still not HTTPS traffic. If you haven't already, I suggest reading the OpenVPN [manpage](https://community.openvpn.net/openvpn/wiki/Openvpn24ManPage) which has a lot of useful information on OpenVPN in TLS mode, tls-auth, and tls-crypt

alexhenrie:
Hi @Giraffe1966, the issue is that the OpenVPN server does not work at all if tls-auth is specified instead of tls-crypt. I don't know why exactly that is, but you can easily see the problem if you try it for yourself.

4s3ti:
Not approving this, tls-auth still needs to be there and pre 2.4 still needs to be provided as an option.
https://github.com/pivpn/pivpn/issues/576

PR #729, opened 2019-05-05
Flush disk writes before reboot

Calvin-L:
Background: I installed PiVPN today for the first time. After allowing this script to reboot my Pi at the end of installation, many of the script's install products were missing (including `/etc/openvpn/easy-rsa/pki/Default.txt` and `/etc/pivpn/INSTALL_USER`). PiVPN struggled along and seemed like it was working, but generated corrupt .ovpn files so I could not set up my client. The install log (which unfortunately got lost after I reinstalled!) did not indicate an error.

I hypothesize (admittedly with very little evidence) that the reboot clobbered some pending writes before they were flushed to my Pi's very very slow SD card.

This commit introduces a `sync` call before the install script exits, which should increase the chance that the writes done during the install survive reboot.


4s3ti:
thanks a lot for your contribution to PiVPN. 
seems reasonable, and won't do any harm for sure.

4s3ti:
Closing this for now... 
will get the master and test properly synced. 

Once I have done it please open a new PR after I have done it. 
I know it's not the best approach here, but I want to avoid issues and conflicts for now.

PR #730, opened 2019-05-07
Fix for OpenVPN version 2.4 check

evolozzy:
Instead of checking OS Version, I made two changes to check OpenVPN version directly. 
- The users who install OpenVPN manually or Debian Buster can use this


4s3ti:
Hi @evolozzy ... thanks a lot for your contribution and help, however, have you tested this? 
I am almost sure this will break the script. and I don't really see the use here... 
OS version check is still needed .. and support for the previous version is still needed as well. 

can you elaborate why is this beneficial?

evolozzy:
Hi @4s3ti,

Sorry it's my bad, will do it.

Issue #696, opened 2019-03-22
why are dots in username disallowed ? 


devZer0:
why is only alphanumeric characters and dashes allowed in pivpn ? please add "." , "_" and "@" so it matches allowed openvpn authentication characters.

we always used "my.name-vpn" for accounts and don't want to change naming scheme. i commented out check routine makeOVPN.sh and it seems to work as usual with "." in name.

so why is dot disallowed?

devZer0:
from openvpn manpage:
<snipp>
      --auth-user-pass-verify cmd method
<snipp>
              To protect against a client passing a maliciously formed username or password string, the username string must consist only of these characters: alphanumeric, underbar ('_'),  dash  ('-'),  dot
              ('.'),  or  at  ('@').   The  password string can consist of any printable characters except for CR or LF.  Any illegal characters in either the username or password string will be converted to
              underbar ('_').
<snipp>

devZer0:
for example, i would like to use email adresses for the login-name, to have easy mechanism to send messages to vpn users based on openvpn log. 

so, could you please change makeOVPN.sh to contain this one ?

if [[ "${NAME}" =~ [^a-zA-Z0-9_@\-\.] ]]; then
    echo "Name can only contain alphanumeric characters and dashes (-)."
    exit 1
fi

devZer0:
to be exact - it's not the username but the certificate name/common name in subject field of certificate - and there is no special restriction with that, too:

https://stackoverflow.com/questions/5136198/what-strings-are-allowed-in-the-common-name-attribute-in-an-x-509-certificate

Giraffe1966:
@devZer0 Not really sure, but possibly because if the first character is a dot then the file will become hidden, possibly confusing users who've never heard of hidden files. Of course somebody could add a prompt disallowing or discouraging names that start with a dot, but I don't think anybody thought of this or had the motivation to code it. Perhaps you could add it in?

4s3ti:
Putting this up for investigation ... Might be worth looking at it ... 

openvpn user is not a system user by itself so it should not be a problem adding it ... maybe @Giraffe1966 is right ... feel free to investigate it if you want @devZer0

4s3ti:
already available in test branch. 
to be closed when test gets merged into master.

4s3ti:
this has been already pushed into master with PR #740 
Closing this issue. 

Thank you for your interest.

Issue #530, opened 2018-05-04
Certificate valid period and auto login


vikingza:
Good day,

I cannot see how to set the valid period for the user so I can either set their login validation for 10 minutes, 1 year or to never expire, I am looking at using openvpn on a raspberry pi for access to a Search and Rescue server which we use for logging of deployments, but I want to be able to set how long the user may have access for before he needs to apply for his / her profile to be renewed (re-activated).

I am also cant find how to make a opvn file to remember its password so that each time one connects, the user does not have to put in a password, I am looking at using this on autonomous systems like field servers that one just connects to a router with internet and Wifi access before powering up, once powered up it can auto "dial" the OpenVPN gateway and then start syncing the field databases with the main servers databases.

4s3ti:
@vikingza sorry for the late reply the project has been stalled for quite a while and we are trying to get back on track now. 

Get certificates to auto-expire seems quite a cool feature however i don't really know how to make it happen, @orazioedoardo , @devZer0 what you think about this? 

> am also cant find how to make a opvn file to remember its password 

I believe that is an option at the User side and it might depend on the client that is being used, not something the server admin can fix on his side. 
however, I highly discourage this for the obvious security reasons.

orazioedoardo:
@4s3ti You can do that by tweaking the EASYRSA_CERT_EXPIRE variable when creating a new client. [By default](https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/vars.example#L124) it's 1080 days, about 3 years.

Giraffe1966:
@vikingza About not typing in a password, you could just run `pivpn add nopass` to generate a private key that isn't encrypted with a password, or use a client feature to remember your password (setting it up depends on the client. It is unwise in high-security environments, as it is essentially the same security as no password at all).

4s3ti:
already available in test brach.
Issue to be closed after test gets merged into master.

4s3ti:
this has been already pushed into master with PR #740
Closing this issue.

Thank you for your interest.

PR #731, opened 2019-05-08
Custom certificate duration and more flexible names

orazioedoardo:
As per request on issue #530 I added the ability to make client certificates last a custom number of days. 
I also relaxed the check for characters in the username as suggested in #696.
@Giraffe1966 pointed out that clients starting with dot would be hidden, so I chose just not to allow them.
Similarly, names starting with dash should also not be allowed since they would be interpreted as command line arguments when provided via... command line argument.


4s3ti:
Seems alright to me... I've tested the test branch everything seems quite right, merging this PR into test. 
Thanks @orazioedoardo

PR #736, opened 2019-05-13
Merge test into master

orazioedoardo:


4s3ti:
This is the opposite of what i was trying to do ... but also works ... 
I ran the test branch a couple of times and it worked perfectly ... 
@james-lasersoft are we good to merge test into master?

4s3ti:
After looking to the diffs everything looks pretty ok, only one questions, 

why is there hardcoded /24 subnet in setDHCPCD function? here https://github.com/pivpn/pivpn/pull/736/files#diff-299d9b30c952310aace4aba77c698874R338

That will certainly break the network of users with other types of subnets.

auto_install/install.sh
```diff
@@ -340,7 +338,7 @@ It is also possible to use a DHCP reservation, but if you are going to do that,
 setDHCPCD() {
     # Append these lines to dhcpcd.conf to enable a static IP
     echo "interface ${pivpnInterface}
-    static ip_address=${IPv4addr}
+    static ip_address=${IPv4addr}/24
```
4s3ti: Hardcoding /24 subnets doesn't seem to be a correct approach.

auto_install/install.sh
```diff
@@ -352,7 +350,7 @@ setStaticIPv4() {
             echo "::: Static IP already configured."
         else
             setDHCPCD
-            $SUDO ip addr replace dev "${pivpnInterface}" "${IPv4addr}"
+            $SUDO ip addr replace dev "${pivpnInterface}" "${IPv4addr}/24"
```
4s3ti: same here, hardcoded /24

auto_install/install.sh
orazioedoardo: I think it's my fault. Basically I added that in my fork because setting a static IP on the raspberry without (my) netmask actually used a /8 subnet. So yes, perhaps there should be some code that detects the netmask or just revert to the default.

4s3ti:
not merging.. will close .. remove the hardcoded /24 , then make a merge from test to master.

PR #740, opened 2019-05-13
Merging Test Branch into Master

4s3ti:
This PR has all conflicts between master and test solved. 
Merges Test Branch into Master. 

If something gets broken ... well .. flame me! =D


4s3ti:
@orazioedoardo @james-lasersoft @Giraffe1966 @devZer0 ... looks clean now, Can you just double check before hitting the Merge button? 

Thanks! =)

orazioedoardo:
Looks good to me, except the issue template. Basically I removed the request for the output of iptables and netstat because the debug script does those checks. Should we keep them for redundancy?

4s3ti:
@orazioedoardo you are right, missed that. if debug script is doing it already I don't think there is a need for it. Will remove that chunk then.

4s3ti:
Done
Removed the IPTables and netstat lines from issue template. 
Merging test branch into master.

PR #773, opened 2019-06-20
.ovpn12 files

IcedComputer:
Added new step to create an .ovpn12 file that can be stored on iOS keychain
This step is more secure method and does not require the end-user to keep entering passwords, or storing the client private cert where it can be easily tampered based on documentation located:
 https://openvpn.net/faq/how-do-i-use-a-client-certificate-and-private-key-from-the-ios-keychain/

Someone can improve upon this by adding a parameter (possibly -i|--iOS) and then generating the original .ovpn file to not contain the client private certificate.

### DIFF
--- a/scripts/makeOVPN.sh
+++ b/scripts/makeOVPN.sh
@@ -281,7 +281,7 @@ if [ "$iOS" = "1" ]; then
 	printf "========================================================\n"
 	openssl pkcs12 -passin pass:$PASSWD -passin pass:$PASSWD -export -in issued/${NAME}${CRT} -inkey private/${NAME}${KEY} -certfile ${CA} -name ${NAME} -out /home/$INSTALL_USER/ovpns/$NAME.ovpn12
 	chown "$INSTALL_USER" "/home/$INSTALL_USER/ovpns/$NAME.ovpn12"
-	chmod o-r "/home/$INSTALL_USER/ovpns/$NAME.ovpn12"
+	chmod 600 "/home/$INSTALL_USER/ovpns/$NAME.ovpn12"
 	printf "========================================================\n"
 	printf "\e[1mDone! %s successfully created!\e[0m \n" "$NAME.ovpn12"
 	printf "You will need to transfer both the .ovpn and .ovpn12 files\n"
--- a/scripts/makeOVPN.sh
+++ b/scripts/makeOVPN.sh
@@ -268,6 +268,17 @@ echo "tls-auth Private Key found: $TA"
 
 } > "${NAME}${FILEEXT}"
 
+
+## Added new step to create an .ovpn12 file that can be stored on iOS keychain
+## This step is more secure method and does not require the end-user to keep entering passwords, or storing the client private cert where it can be easily tampered
+## https://openvpn.net/faq/how-do-i-use-a-client-certificate-and-private-key-from-the-ios-keychain/
+printf "========================================================\n"
+printf "Generating an .ovpn12 file for use with iOS devices\n"
+printf "You will be prompted to re-enter some information from the cert you just created\n"
+printf "========================================================\n"
+
+sudo openssl pkcs12 -export -in issued/${NAME}${CRT} -inkey private/${NAME}${KEY} -certfile ${CA} -name ${NAME} -out /home/$INSTALL_USER/ovpns/$NAME.ovpn12
+
 # Copy the .ovpn profile to the home directory for convenient remote access
 cp "/etc/openvpn/easy-rsa/pki/$NAME$FILEEXT" "/home/$INSTALL_USER/ovpns/$NAME$FILEEXT"
 chown "$INSTALL_USER" "/home/$INSTALL_USER/ovpns/$NAME$FILEEXT"
--- a/scripts/makeOVPN.sh
+++ b/scripts/makeOVPN.sh
@@ -274,10 +274,12 @@ echo "tls-auth Private Key found: $TA"
 ## https://openvpn.net/faq/how-do-i-use-a-client-certificate-and-private-key-from-the-ios-keychain/
 printf "========================================================\n"
 printf "Generating an .ovpn12 file for use with iOS devices\n"
-printf "You will be prompted to re-enter some information from the cert you just created\n"
 printf "========================================================\n"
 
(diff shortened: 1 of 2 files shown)


4s3ti:
Will let other most active users comment on this as I am not at all familiar with IOS devices. 

If their feedback is positive then ill gladly merge it. 

@orazioedoardo @devZer0 @Giraffe1966  any takes on this?

orazioedoardo:
Yeah, I see that this is more secure because the private parts of the config are not in the .ovpn file, though it's a bit less convenient since the user need to copy two files, both the .ovpn and the .ovpn12 and instruct the user on how to use them. Also your script doesn't do the whole job:
- the .ovpn12 file is root owned in the ovpn folder so you should chown and chmod it to allow the user to read it.
- you need to strip cert and key sections (or just don't add them) in the .ovpn otherwise you defeat initial purpose.
- probably you should use expect to type the export password and the private key password non-interactively.

By the way I wouldn't make this the default, I would add a one time dialog after the first client creation with some text like "TIP: If you have an iOS device you can pass -i|--iOS to [...]"

Giraffe1966:
No need to use expect for the password, you can use the `-passin env:PASSWD` and `-passout env:PASSWD` options to openssl to do this non-interactively. Also, don't use sudo for openssl as this is run as root and this would break on systems without sudo installed.

IcedComputer:
Thanks for the feedback.  I will be sure to incorporate it.


> * you need to strip cert and key sections (or just don't add them) in the .ovpn otherwise you defeat initial purpose.

> By the way I wouldn't make this the default, I would add a one time dialog after the first client creation with some text like "TIP: If you have an iOS device you can pass -i|--iOS to [...]"

I am a bit new at this, so trying to figure out how to use the -i|--iOS filter correctly to create this and strip out the cert and key sections.  I will do some testing and upload something that works!

IcedComputer:
Updated.

4s3ti:
@orazioedoardo , @Giraffe1966  what is your take on this? is it good to be merged? 

thanks a lot for keeping the boat moving! =)

scripts/makeOVPN.sh
```diff
@@ -13,14 +13,15 @@ INSTALL_USER=$(cat /etc/pivpn/INSTALL_USER)
...
-    echo "::: Usage: pivpn <-a|add> [-n|--name <arg>] [-p|--password <arg>]|[nopass] [-d|--days <number>] [-h|--help]"
+    echo "::: Usage: pivpn <-a|add> [-n|--name <arg>] [-p|--password <arg>]|[nopass] [-d|--days <number>] [-i|--iOS] [-h|--help]"
     echo ":::"
     echo "::: Commands:"
     echo ":::  [none]               Interactive mode"
     echo ":::  nopass               Create a client without a password"
     echo ":::  -d,--days            Expire the certificate after specified number of days (default: 1080)"
     echo ":::  -n,--name            Name for the Client (default: '"$(hostname)"')"
     echo ":::  -p,--password        Password for the Client (no default)"
+	echo ":::  -i,--iOS             Generate a certificate that leverages iOS keychain"
```
Giraffe1966: Perhaps include a link to https://openvpn.net/faq/how-do-i-use-a-client-certificate-and-private-key-from-the-ios-keychain/ so users know how to use the .ovpn12 file, and why this option increases security.

scripts/makeOVPN.sh
```diff
@@ -235,8 +239,58 @@ if [ ! -f "${TA}" ]; then
...
+	# Copy the .ovpn profile to the home directory for convenient remote access
+		
+	printf "========================================================\n"
+	printf "Generating an .ovpn12 file for use with iOS devices\n"
+	printf "Please remember the export password\n"
+	printf "as you will need this import the certificate on your iOS device\n"
+	printf "========================================================\n"
+	openssl pkcs12 -passin pass:$PASSWD -passin pass:$PASSWD -export -in issued/${NAME}${CRT} -inkey private/${NAME}${KEY} -certfile ${CA} -name ${NAME} -out /home/$INSTALL_USER/ovpns/$NAME.ovpn12
+	chown "$INSTALL_USER" "/home/$INSTALL_USER/ovpns/$NAME.ovpn12"
+	chmod o-r "/home/$INSTALL_USER/ovpns/$NAME.ovpn12"
```
Giraffe1966:
```suggestion
	chmod 600 "/home/$INSTALL_USER/ovpns/$NAME.ovpn12"
```
`chmod 600` only allows read+write for owner (which is probably what most people would want), while `chmod o-r` might still allow a user who is not the owner or in the owning group to write to the file, if this is run on a user's machine with weird default permissions (where random users can write to others' files by default).

scripts/makeOVPN.sh
```diff
@@ -235,8 +239,58 @@ if [ ! -f "${TA}" ]; then
...
+	} > "${NAME}${FILEEXT}"
+	
+	# Copy the .ovpn profile to the home directory for convenient remote access
+		
+	printf "========================================================\n"
+	printf "Generating an .ovpn12 file for use with iOS devices\n"
+	printf "Please remember the export password\n"
+	printf "as you will need this import the certificate on your iOS device\n"
+	printf "========================================================\n"
+	openssl pkcs12 -passin pass:$PASSWD -passin pass:$PASSWD -export -in issued/${NAME}${CRT} -inkey private/${NAME}${KEY} -certfile ${CA} -name ${NAME} -out /home/$INSTALL_USER/ovpns/$NAME.ovpn12
```
Giraffe1966: I believe using `-passin pass:$PASSWD` twice is a typo, and, according to the manpage, using `-passin env:PASSWD` might be more secure.

scripts/makeOVPN.sh
4s3ti: Yup agree with that one there. 
also for sake of consistency and verbosity numbers should always be used.

scripts/makeOVPN.sh
4s3ti: wouldn't a link be too much of information there? 
Maybe we should put the link in the FAQ or wiki or whatsoever and get a simple reference to it?

4s3ti:
@IcedComputer please apply the changes and suggestions mentioned by @Giraffe1966 and ill merge your PR

4s3ti:
Thanks @IcedComputer , Merging.

PR #777, opened 2019-07-01
Hide client IPs inside the debug log

orazioedoardo:
This pull requests redacts all public IP addresses inside the snippet of the server log, which are usually client IPs. It does such by using grepcidr to select lines with non private IP ranges, uses grep and uniq to extract the individual IPs and sed to replace them.


4s3ti:
LGTM .. Thanks a lot for your help @orazioedoardo ... 
Merging.

Issue #774, opened 2019-06-22
vanilla pivpn lan and internet doesnt work


judos:
## In raising this issue, I confirm the following: 

- [x] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*.
- [?] The issue I am reporting can be *is* directly related to the pivpn installer script.
- [x] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pivpn/pivpn/wiki/FAQ), [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen)).

### Issue

I installed pivpn successfully and can connect to it but lan/internet doesn't seem reachable as soon as I do. I tried accessing the router/google and they both don't load.

### Have you searched for similar issues and solutions?
yes: https://github.com/pivpn/pivpn/issues/207, https://github.com/pivpn/pivpn/issues/634, https://github.com/pivpn/pivpn/issues/676, https://github.com/pivpn/pivpn/issues/752

### Console output of      `pivpn debug`
https://pastebin.com/CEbvUQfz

### Have you taken any steps towards solving your issue?

I tried several of the proposed solutions in the other tickets, but unluckily nothing worked so far.
No compression config is in the server.conf, and setting local ip, or push route also didn't seem to work.
Aside this at least I only used pivpn and nothing else, no pihole etc. Just a vanilla install with 2 users added. And everything from the installation to the connect vpn works. Just I don't reach any device then.

Any clues what might be wrong?
Thanks in advance for your help!

orazioedoardo:
The connection seem to be established. When connected to the VPN, can you SSH into your raspberry pi via these addresses: 192.168.1.10 and 10.8.0.1? I see you are on android, you can use [JuiceSSH](https://play.google.com/store/apps/details?id=com.sonelli.juicessh&hl=it).

judos:
Yes the connection seems generally ok in the android OpenVPN Connect client.
Yeah JuiceSSH is very nice, already use that a lot.
I can connect with 192.168.1.10 to the raspberry. (but it warns me that the public key of the raspberry has changed and that the connection might be unsecure therefore. Seems odd to me..)
Also 10.8.0.1 works fine.

[...]

PR #778, opened 2019-07-01
Improve iptables detection

orazioedoardo:
Handle non-default firewall configuration where the user may already have some rules that drop incoming packets or block forwarding, like this one #774.
Also: removed `$SUDO` from the uninstall script since `pivpn` already calls it with sudo.


4s3ti:
LGTM... Merging.

PR #781, opened 2019-07-03
Only use iptables-legacy if platform is Buster

orazioedoardo:
Just noticed this, otherwhise the script will exit on other platforms because of `set -e`.


4s3ti:
Once again, thanks for the great work @orazioedoardo

Issue #640, opened 2018-11-05
Port 1194 not open


johnwaynesn:
Hi,

i have troubles with opening ports on my Raspberry Pi running Raspbian Stretch Version October 2018.

I tried different instruductions of setting up a VPN server on a Raspberry Pi but none of them worked for me. I always have this problem with port opening. The port 1194 is not opened from the local network. I think i am doing something fundamental wrong on the linux iptables setup.

At this time i set up my system from scratch and only installed pivpn using your script. The installation was successful i guess.

As you can see in the very last output below the netstat command shows no entry for the port 1194. So this might be the reason for not being open right? But how do i correctly setting up port opening on my system?

I hope you are willing to help me here out since thats not exactly a pivpn issue.

Thank you in advance and for this wonderful script aswell.

## Console output of `curl -L install.pivpn.io | bash`
```
pi@raspberrypi:~ $ curl -L install.pivpn.io | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   178  100   178    0     0    448      0 --:--:-- --:--:-- --:--:--   450
100 53084  100 53084    0     0  64374      0 --:--:-- --:--:-- --:--:-- 64374
:::
::: sudo will be used for the install.
::: Update option selected.
::: Verifying free disk space...
:::
::: Checking apt-get for upgraded packages.... done!
:::
::: Your system is up to date! Continuing with PiVPN installation...
::: Using IP address: 192.168.1.20
::: Using interface: eth0
::: Using User: pi
::: Using protocol: udp
::: Using port: 1194
:::
::: Static IP already configured.
:::
::: Checking for existing base files...
:::    Checking /etc/.pivpn is a repo... OK!
:::     Updating repo in /etc/.pivpn... done!
:::
::: Stopping OpenVPN service... done.
:::
::: Installing scripts to /opt/pivpn... done.
set_var EASYRSA_ALGO       ec
set_var EASYRSA_CURVE
::: Building CA...
unknown curve name ()

Easy-RSA error:

Curve  not found. Run openssl ecparam -list_curves to show a
list of supported curves.
::: Restarting services...
Synchronizing state of openvpn.service with SysV service script with /lib/systemd/systemd-sysv-install.
Executing: /lib/systemd/systemd-sysv-install enable openvpn
::: done.
:::
::: Update complete!
:::
::: The install log is located at: /etc/pivpn/install.log
```

[...]

Issue #787, opened 2019-07-11
OpenVPN service does not run / allow client to connect


Pysis868:
## In raising this issue, I confirm the following: 

`{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting can be *is* directly related to the pivpn installer script.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pivpn/pivpn/wiki/FAQ), [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen)).

### Issue
Using my configuration route, the server is not running/listening.

### Have you searched for similar issues and solutions?
    yes, I have only found the correct scenario represented in some logs, but I have a problem trying getting that far.
    Also, maybe many do not choose 4096 key size, so finding my exact .pem string and situation to search for does not find many issues either.

### Console output of      `curl -L install.pivpn.io | bash`
```
pi@pivpn:~ $ curl -L install.pivpn.io | bash 2>&1
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   178  100   178    0     0    681      0 --:--:-- --:--:-- --:--:--   684
100 55086  100 55086    0     0   126k      0 --:--:-- --:--:-- --:--:--  126k
:::
::: sudo will be used for the install.
::: Did not detect perfectly supported OS but,
::: Continuing installation at user's own risk...
::: Update option selected.
::: Verifying free disk space...
:::
::: Checking apt-get for upgraded packages.... done!
:::
::: Your system is up to date! Continuing with PiVPN installation...
::: Using IP address: ...
::: Using interface: ...
::: Using User: ...
::: Using protocol: ...
::: Using port: ...
:::
::: Static IP already configured.
:::
::: Checking for existing base files...
:::    Checking /etc/.pivpn is a repo...:::    Cloning https://github.com/pivpn/pivpn.git into /etc/.pivpn... done!
:::
::: Stopping OpenVPN service... done.
:::
::: Installing scripts to /opt/pivpn... done.
[...]

Issue #786, opened 2019-07-06
- is no longer allowed in client 


jpknz:
Using `pivpn -a nopass -n name-here -d 365` the response recieved is `Name can only contain alphanumeric characters and these characters (.-@_).`

`@` and `_` work as expected but `-` gives the above error

orazioedoardo:
Thanks, apparently, there is a missing `\` before `-` [here](https://github.com/pivpn/pivpn/blob/8e3a95152412a9d35b67d51bfc80379350815252/scripts/makeOVPN.sh#L150).

tycrek:
This error occurs in interactive mode as well: `pivpn add nopass`

mquirin:
> Thanks, apparently there is a missing `\` right before the fourth `-` [here](https://github.com/pivpn/pivpn/blob/8e3a95152412a9d35b67d51bfc80379350815252/scripts/makeOVPN.sh#L150).

I think this works differently than you intend: After patching makeOVPN.sh with the change
you propose it also accepts '\' as a valid char (e.g. 'foo\bar' is not refused).

I think a better pattern is this: `[^a-zA-Z0-9.@_-]`

Have a look here as well (according to this there is no real escaping within character classes in Posix ERE (which IIRC is what bash is using)):
https://stackoverflow.com/questions/399078/what-special-characters-must-be-escaped-in-regular-expressions

orazioedoardo:
Thanks for the insights @mquirin, so since `-` is a metacharacter used to represent a range, you have to place it just after `^` or as the last character if you want the shell to interpret it as a literal `-`.

4s3ti:
Test branch merged into master. 
Closing

Issue #784, opened 2019-07-05
Question : Client Cert # of days until expiry


jpknz:
This is a question rather than a bug/issue. I've just noticed that on a new install of PiVPN when creating a new client is slightly different in that it asks for the number of days before expiry whereas before it didn't. I couldn't find any info on this so am asking here. 

What was the default expiry previously (if any)?
Why was just under 3 years chosen as the expiry period?

Thanks

orazioedoardo:
> I've just noticed that on a new install of PiVPN when creating a new client is slightly different in that it asks for the number of days before expiry whereas before it didn't. I couldn't find any info on this so am asking here.

See https://github.com/pivpn/pivpn/pull/731.

> What was the default expiry previously (if any)?

I think 10 years.

> Why was just under 3 years chosen as the expiry period?

It's just the [default value](https://github.com/OpenVPN/easy-rsa/blob/6b7b6bf1f0d3c9362b5618ad18c66677351cacd1/easyrsa3/vars.example#L126) for easy-rsa if you don't set one.

orazioedoardo:
Oh wait, now you got me thinking: given that by default certificates last for 1080 days, this means that the server certificate created at the time of installation lasts just as long. For this reason setting a client expiration period of over 1080 days is useless since the server expires sooner! I might send a pull request later next week...

jpknz:
Ahh thanks.
Think I created my server cert outside of pivpn anyway and set it to 10 years (don't remember the exact process but it was directly using easy-rsa I think). I just use PiVpn for the easy add/remove client scripts

4s3ti:
test branch merged into master. 
closing.

PR #788, opened 2019-07-13
Miscellaneous fixes

orazioedoardo:
- Issue a server cert that lasts 3650 days instead of the default 1080, so it won't expire before client certs that last more than 1080 days: https://github.com/pivpn/pivpn/issues/784#issuecomment-508844437
- Actually allow dashes in usernames (@mquirin): https://github.com/pivpn/pivpn/issues/786

- The script called `update_repo` [without the second argument](https://github.com/pivpn/pivpn/blob/8e3a95152412a9d35b67d51bfc80379350815252/auto_install/install.sh#L531), the pivpn git url, making `git clone [...]` fail. Moreover, since `/etc/.pivpn`, the current working directory, is deleted [here](https://github.com/pivpn/pivpn/blob/8e3a95152412a9d35b67d51bfc80379350815252/auto_install/install.sh#L547), git will complain, so we move back to /etc: https://github.com/pivpn/pivpn/issues/787

```
::: Checking for existing base files...
:::    Checking /etc/.pivpn is a repo... OK!
:::     Updating repo in /etc/.pivpn... [/]  fatal: repository '' does not exist
./install.sh: line 565: cd: /etc/.pivpn: No such file or directory
```
- The mapping from ecdsa cert size and curve name is [declared](https://github.com/pivpn/pivpn/blob/8e3a95152412a9d35b67d51bfc80379350815252/auto_install/install.sh#L782) inside [this if statement](https://github.com/pivpn/pivpn/blob/8e3a95152412a9d35b67d51bfc80379350815252/auto_install/install.sh#L752), that is not reached when the user installs pivpn over an existing installation (update option). The result is that EASYRSA_CURVE is undefined. We move the associative array outside the if statement to fix: #640

```
+ echo 'set_var EASYRSA_ALGO       ec'
+ tee -a vars
set_var EASYRSA_ALGO       ec
+ echo 'set_var EASYRSA_CURVE      '
+ tee -a vars
set_var EASYRSA_CURVE
+ ./easyrsa --batch init-pki
+ printf '::: Building CA...\n'
+ ./easyrsa --batch build-ca nopass
::: Building CA...
unknown curve name ()

Easy-RSA error:

Curve  not found. Run openssl ecparam -list_curves to show a
list of supported curves.
```
- Make grep return true when there are no matching iptables rules, to avoid conflicts with `set -e`.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -565,6 +565,8 @@ update_repo() {
         # Pull the latest commits
         echo -n ":::     Updating repo in $1..."
         $SUDO rm -rf "${1}"
+        # Go back to /etc otherwhise git will complain when the current working directory has
+        # just been deleted (/etc/.pivpn).
         cd /etc
         $SUDO git clone -q --depth 1 --no-single-branch "${2}" "${1}" > /dev/null & spinner $!
         cd "${1}" || exit 1
@@ -754,6 +756,8 @@ confOpenVPN() {
     NEW_UUID=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 16 | head -n 1)
     SERVER_NAME="server_${NEW_UUID}"
 
+    declare -A ECDSA_MAP=(["256"]="prime256v1" ["384"]="secp384r1" ["521"]="secp521r1")
+
     if [[ ${useUpdateVars} == false ]]; then
         # Ask user for desired level of encryption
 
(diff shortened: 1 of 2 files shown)


4s3ti:
LGTM Great work, and thanks a lot @orazioedoardo merging to test... as soon as I get back from my vacation and I have a couple of mins I will try the test branch and see how it's working to merge it into master.

PR #822, opened 2019-09-02
Fix .ovpn12 file, make pivpn -a options discoverable

orazioedoardo:
`-passin env:$PASSWD` reads the password from the `$PASSWD` environment variable, but `$PASSWD` is not an environment variable, it's the actual password. Now using `-passin pass:"$PASSWD"` to fix.

`--bitwarden` option is shown both when running `pivpn` and running `pivpn -a` which seems redundant to me, so I removed it from `pivpn`. However, to make it discoverable, together with `--iOS` and `--days`, we show the client creation options the first time the user runs `pivpn -a`.

### DIFF
--- a/pivpn
+++ b/pivpn
@@ -58,7 +58,6 @@ function helpFunc {
     echo ":::"
     echo "::: Commands:"
     echo ":::  -a, add [nopass]     Create a client ovpn profile, optional nopass"
-    echo ":::  -b,--bitwarden       Create and save a client through Bitwarden"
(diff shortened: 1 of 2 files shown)


4s3ti:
Merging, 
Soon will be pushing the fix for bitwarden failing to install as well as making it optional.

4s3ti:
@orazioedoardo almost merged before the changes! =D 
can i merge it now?

orazioedoardo:
> @orazioedoardo almost merged before the changes! =D

LOL, true. Yes you can.

Issue #829, opened 2019-09-05
Subdomain not allowed in custom dns search list


DirkLaurenz:
## In raising this issue, I confirm the following: 

`{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting can be *is* directly related to the pivpn installer script.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pivpn/pivpn/wiki/FAQ), [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen)).

Entering a custom domain refuses subdomains. In my case i have an officially registered domain example.com and i have a local samba based AD (samba.example.com). I Provided the samba dns servers and want to provide a custom dns search domain (samba.example.com) but i can only enter example.com - after install i can change to samba.example.com manually...
setup itself runs perfectly, vpn works fine

### Have you searched for similar issues and solutions?
 yes

### Console output of      `curl -L install.pivpn.io | bash`
root@vpn01:~# curl -L https://install.pivpn.io | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   178  100   178    0     0     93      0  0:00:01  0:00:01 --:--:--    93
100 58458  100 58458    0     0  20569      0  0:00:02  0:00:02 --:--:--  151k
:::
::: You are root.
::: Verifying free disk space...
:::
::: Checking apt-get for upgraded packages.... done!
:::
::: Your system is up to date! Continuing with PiVPN installation...
update-alternatives: using /usr/sbin/iptables-legacy to provide /usr/sbin/iptables (iptables) in manual mode
update-alternatives: using /usr/sbin/ip6tables-legacy to provide /usr/sbin/ip6tables (ip6tables) in manual mode
::: Using User: adminl
:::
::: Checking for existing base files...
:::    Checking /etc/.pivpn is a repo...:::    Cloning https://github.com/pivpn/pivpn.git into /etc/.pivpn... done!
:::
::: Stopping OpenVPN service... done.
:::
[...]

PR #848, opened 2019-10-14
Allow subdomains in custom dns search list

rrobgill:
Addresses #829 , can specify subdomains. 

Regex specifies : must start with alphanumeric, may have no more
than two consecutive hyphens, must not end or start with hyphen,
final word must be at least two apha characters.

Signed-off-by: Rob Gill <rrobgill@protonmail.com>


zkhan96MSM:
fantastic work guys!!

PR #863, opened 2019-11-07
Add uninstall script and Pi-hole detection

orazioedoardo:
This pull requests brings back the uninstall script, which means that `pivpn -u` works again. The uninstall script will only prompt to uninstall packages that where not already installed at the time of the PiVPN installation. `installDependentPackages()` will add pending packages to the `TO_INSTALL` array that is saved to `setupVars.conf` for later referencing when uninstalling PiVPN.

The installer now detects and offer to use Pi-hole as the DNS server for the VPN.


4s3ti:
kewl!

PR #873, opened 2019-12-07
add missing bang to shebang. correct usage of cp. correct wrong copy …

corbolais:
…target dir. safeguard sudo-rm-rf. safeguard against variable expansion into IFS chars. clean up whitespace.

Signed-off-by: corbolais <corbolais@gmail.com>

### DIFF
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -1,40 +1,40 @@
-#/bin/bash
+#!/bin/bash
 
 ###Updates pivpn scripts (Not PiVPN)
 ###Main Vars
 pivpnrepo="https://github.com/pivpn/pivpn.git"
 pivpnlocalpath="/etc/.pivpn"
-pivpnscripts="/opt/pivpn/scripts"
-bashcompletiondir="/etc/bash_completion.d/pivpn"
+pivpnscripts="/opt/pivpn/"
+bashcompletiondir="/etc/bash_completion.d/"
 
 
 ###Functions
 ##Updates scripts
 updatepivpnscripts(){
-    ##We don't know what sort of changes users have made. 
+    ##We don't know what sort of changes users have made.
     ##Lets remove first /etc/.pivpn dir then clone it back again
     echo "going do update PiVPN Scripts"
-    if [[ -d $pivpnlocalpath ]]; then
-
-        sudo rm -rf $pivpnlocalpath
+    if [[ -d "$pivpnlocalpath" ]]; then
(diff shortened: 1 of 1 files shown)


4s3ti:
@corbolais Thanks a lot for this, however i have to Deny this one, 
1 - No changes are pushed directly to master,
2 - We are locking down Master branch as Test has a lot of new code and changes that will after take over the master branch, you should pull that one and push those fixes there if they still exist. 

Once again, sorry for this, and thanks a lot for helping to improve pivpn.

orazioedoardo:
Agree with @4s3ti. @corbolais please make a new pull request using test as reference. Your pull requests overlap too, one is enough for the changes you suggested.

Good point on [this](https://github.com/pivpn/pivpn/pull/874/files#diff-299d9b30c952310aace4aba77c698874L551-R568) and more safeguards, however I'm not a fan of supporting experimental distro releases (the test repo has been tested on debian buster and ubuntu bionic).

corbolais:
fair enough.
my pain.

corbolais:
update.sh got refactored or just removed?

corbolais:
> experimental distro releases

buster isn't really an experimental distro. fact ist, I'd like to use pivpn on bullseye/sid. I provided the required changes to have it so. Why not merge into test then.

orazioedoardo:
> update.sh got refactored or just removed?

Still there, just moved: https://github.com/pivpn/pivpn/blob/test/scripts/openvpn/update.sh

> buster isn't really an experimental distro.

Sorry I meant that the script has not been tested on experimental releases, only on stable releases like buster and bionic.

> fact ist, I'd like to use pivpn on bullseye/sid.

You can still use it, even though the script will tell you that the relase may not work.

> I provided the required changes to have it so. Why not merge into test then.

My rationale is that if we claim that pivpn has been tested on bullseye/sid, we will need to keep testing the script over and over since the repo may introduce breaking changes.

corbolais:
> My rationale is that if we claim that pivpn has been tested on bullseye/sid, we will need to keep testing the script over and over since the repo may introduce breaking changes.

Same once the releases are finalized and out. But then it will be supported from the get go.

orazioedoardo:
> Same once the releases are finalized and out. But then it will be supported from the get go.

We don't regularly check the script (I mean, on a time basis), only when pushing commits to master, considering how many people work on the project as of now.
Regarding the test branch, I've tested it against the releases above (which do not cover all releases from the master branch as I didn't have time these days).

PR #877, opened 2019-12-08
debian openvpn has dedicated user and group, so rather use those than…

corbolais:
… nobody:nogroup (as do _many_ other packages and thus all having access to such owned files).

Signed-off-by: corbolais <corbolais@gmail.com>

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1294,7 +1294,7 @@ set_var EASYRSA_KEY_SIZE   ${pivpnENCRYPT}" | $SUDO tee vars >/dev/null
 	${SUDOE} cp pki/crl.pem /etc/openvpn/crl.pem
   if test "${PLAT}" = "Debian"; then
 		if ! getent passwd openvpn; then
-			${SUDOE} adduser --system --home /etc/openvpn/ --no-create-home --group ${debianOvpnUserGroup%:*}
+			${SUDOE} adduser --system --home /var/lib/openvpn/ --no-create-home --group --disabled-login ${debianOvpnUserGroup%:*}
 		fi
 		${SUDOE} chown "$debianOvpnUserGroup" /etc/openvpn/crl.pem
   else
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -14,6 +14,8 @@
 setupVars=/etc/pivpn/setupVars.conf
 pivpnFilesDir="/etc/.pivpn"
 
+debianOvpnUserGroup="openvpn:openvpn"
+
 ### PKG Vars ###
 PKG_MANAGER="apt-get"
 PKG_CACHE="/var/lib/apt/lists/"
@@ -1290,7 +1292,11 @@ set_var EASYRSA_KEY_SIZE   ${pivpnENCRYPT}" | $SUDO tee vars >/dev/null
 	# Generate an empty Certificate Revocation List
 	${SUDOE} ./easyrsa gen-crl
 	${SUDOE} cp pki/crl.pem /etc/openvpn/crl.pem
-	${SUDOE} chown nobody:nogroup /etc/openvpn/crl.pem
+  if test "${PLAT}" = "Debian"; then
+		${SUDOE} chown "$debianOvpnUserGroup" /etc/openvpn/crl.pem
+  else
+		${SUDOE} chown nobody:nogroup /etc/openvpn/crl.pem
+  fi
 
 	# Write config file for server using the template.txt file
 	$SUDO cp /etc/.pivpn/server_config.txt /etc/openvpn/server.conf
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1292,14 +1292,10 @@ set_var EASYRSA_KEY_SIZE   ${pivpnENCRYPT}" | $SUDO tee vars >/dev/null
 	# Generate an empty Certificate Revocation List
 	${SUDOE} ./easyrsa gen-crl
 	${SUDOE} cp pki/crl.pem /etc/openvpn/crl.pem
-  if test "${PLAT}" = "Debian"; then
-		if ! getent passwd openvpn; then
-			${SUDOE} adduser --system --home /var/lib/openvpn/ --no-create-home --group --disabled-login ${debianOvpnUserGroup%:*}
-		fi
-		${SUDOE} chown "$debianOvpnUserGroup" /etc/openvpn/crl.pem
-  else
-		${SUDOE} chown nobody:nogroup /etc/openvpn/crl.pem
+  if ! getent passwd openvpn; then
+    ${SUDOE} adduser --system --home /var/lib/openvpn/ --no-create-home --group --disabled-login ${debianOvpnUserGroup%:*}
   fi
+  ${SUDOE} chown "$debianOvpnUserGroup" /etc/openvpn/crl.pem
 
 	# Write config file for server using the template.txt file
 	$SUDO cp /etc/.pivpn/server_config.txt /etc/openvpn/server.conf
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1293,6 +1293,9 @@ set_var EASYRSA_KEY_SIZE   ${pivpnENCRYPT}" | $SUDO tee vars >/dev/null
 	${SUDOE} ./easyrsa gen-crl
 	${SUDOE} cp pki/crl.pem /etc/openvpn/crl.pem
   if test "${PLAT}" = "Debian"; then
+		if ! getent passwd openvpn; then
+			${SUDOE} adduser --system --home /etc/openvpn/ --no-create-home --group ${debianOvpnUserGroup%:*}
+		fi
 		${SUDOE} chown "$debianOvpnUserGroup" /etc/openvpn/crl.pem
   else
 		${SUDOE} chown nobody:nogroup /etc/openvpn/crl.pem
--- a/scripts/uninstall.sh
+++ b/scripts/uninstall.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 # PiVPN: Uninstall Script
 
+### FIXME: global: config storage, refactor all scripts to adhere to the storage
+### FIXME: use variables where appropriate, reduce magic numbers by 99.9%, at least.
+
 PKG_MANAGER="apt-get"
 setupVars="/etc/pivpn/setupVars.conf"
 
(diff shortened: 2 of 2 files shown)


orazioedoardo:
I don't have the openvpn group on Debian:
```
$ dpkg -s openvpn | grep '^Status:'
Status: install ok installed
$ grep openvpn /etc/group
$ 
```

corbolais:
> I don't have the openvpn user nor group on Debian:

that's interesting. would you mind posting `cat /etc/os-release`?

I've got a couple of opvn servers here, all of them feature the user/group combo.

orazioedoardo:
```
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

corbolais:
Curious
Seems to be a bullseye/sid thing, too.
All openvpn servers here are bullseye/sid:

```
PRETTY_NAME="Debian GNU/Linux bullseye/sid"
NAME="Debian GNU/Linux"
ID=debian
```

```
getent passwd openvpn
openvpn:x:123:130::/var/lib/openvpn:/usr/sbin/nologin
```

orazioedoardo:
Same on Raspbian by the way, can you check Ubuntu too?

corbolais:
```
root:~# grep openvpn /etc/passwd
nm-openvpn:x:124:136:NetworkManager OpenVPN,,,:/var/lib/openvpn/chroot:/usr/sbin/nologin
root:~# cat /etc/os-release 
NAME="Ubuntu"
VERSION="19.04 (Disco Dingo)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 19.04"
VERSION_ID="19.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=disco
UBUNTU_CODENAME=disco
```

Do you want me to add this user to the PLAT=Ubuntu path?

orazioedoardo:
> Do you want me to add this user to the PLAT=Ubuntu path?

Isn't that user related just to the Network Manager integration with openvpn? I would just create an 'openvpn' user for every platform, so you don't need [this test](https://github.com/pivpn/pivpn/pull/877/files#diff-299d9b30c952310aace4aba77c698874R1295) (by the way no nm-openvpn nor openvpn user on ubuntu 18.04).

orazioedoardo:
Or maybe check if ‘nm-openvpn’ exists. If it exists, use it, otherwhise create ‘openvpn’. But what would happen if the user uninstalls or replaces Network Manager? It will remove the ‘nm-openvpn’ user thus breaking the openvpn server.

corbolais:
> so you don't need [this test](https://github.com/pivpn/pivpn/pull/877/files#diff-299d9b30c952310aace4aba77c698874R1295) (by the way no nm-openvpn nor openvpn user on ubuntu 18.04).

That would be my favourite, yes.

orazioedoardo:
Can you add some code to remove the openvpn user if the user chooses to uninstall the openvpn package (in the uninstall.sh script)?

orazioedoardo:
Regarding [this](https://github.com/pivpn/pivpn/pull/877/commits/729674595bb9071fcbf06d4384c72fc456b4dcd4#diff-409efba6db37b831b75997c11a4c9ec4R111), maybe use different naming when creating those files at the time of installation? What's the problem [here](https://github.com/pivpn/pivpn/pull/877/commits/729674595bb9071fcbf06d4384c72fc456b4dcd4#diff-409efba6db37b831b75997c11a4c9ec4R138)?

4s3ti:
Sorry i haven't read the whole topic ... quite tight on time ... and almost finishing with the website in the new domain .. however ... not entirely sure if its related or not wanted to give an heads up. 

Openvpn Drops permissions to nobody:Nogroup due to security reasons .. please be careful to don't go against it!

orazioedoardo:
@4s3ti Actually he's right. Dropping permissions for security reasons is ok but it's effectiveness is reduced if other processes drop to the same users. Makes sense to drop permissions to a specific user only used by openvpn.

4s3ti:
@orazioedoardo aight!! i kinda just saw the notification and wanted to give a heads up first! =) 
it has been tight on time lately but soon will get back on code and be able to allocate regular time to it! =)

corbolais:
> Regarding [this](https://github.com/pivpn/pivpn/pull/877/commits/729674595bb9071fcbf06d4384c72fc456b4dcd4#diff-409efba6db37b831b75997c11a4c9ec4R111), maybe use different naming when creating those files at the time of installation? 

Yes, that's the way to go. Didn't modify on short notice but didn't want to let it pass uncommented either. So, solution to this would be a proper naming scheme. Which IMO falls right in line with a refactor of commonly used functions, magic numbers, general double quoting when using i.e. whiptail and coding style (line spacing and indentation levels/chars in particular). Then again, the naming scheme must be consistent for install and uninstall routines, again, requiring some persistant config storage.

> What's the problem [here](https://github.com/pivpn/pivpn/pull/877/commits/729674595bb9071fcbf06d4384c72fc456b4dcd4#diff-409efba6db37b831b75997c11a4c9ec4R138)?

Basically the same issue: locally user-modified files unconditionally removed. Either use dedicated pipvpn config files where feasible or really only remove stuff pivpn added.

Same goes for iptables/ufw rules. I already improved on this in install.sh regarding in-place modification and forced *nat table addition. Generally, it is recommended to just avoid in-place modification using sed, it is way too fragile and you don't want to mess up users' configs, even less with fw configs in particular.

orazioedoardo:
> Basically the same issue: locally user-modified files unconditionally removed. Either use dedicated pipvpn config files where feasible or really only remove stuff pivpn added.

Agree. It's a bit mitigated by the fact that those files are removed only if pivpn itself installed unattended-upgrades (see [here](https://github.com/pivpn/pivpn/blob/4466f1503c9797808d49d8cb55063bec5a79deed/auto_install/install.sh#L27-L30)).

This could be further expanded by copying pivpn specific setting from files in the git repo instead of inlining them into the script, what do you think?

corbolais:
For what it's worth, maybe it should be considered globally implementing the use of `install(1)`.

corbolais:
> Agree. It's a bit mitigated by the fact that those files are removed only if pivpn itself installed unattended-upgrades (see [here](https://github.com/pivpn/pivpn/blob/4466f1503c9797808d49d8cb55063bec5a79deed/auto_install/install.sh#L27-L30)).

I'm having a hard time seeing this as an effective mitigation with respect to the underlying issue.

> 
> This could be further expanded by copying pivpn specific setting from files in the git repo instead of inlining them into the script, what do you think?

Absolutely.

orazioedoardo:
> I'm having a hard time seeing this as an effective mitigation with respect to the underlying issue.

If the issue is "don't remove files that the user may have already modified for its own purposes" then it's a mitigation, since if pivpn installed the package those files where not there, so it's safe to remove them.

> For what it's worth, maybe it should be considered globally implementing the use of `install(1)`.

If you are willing to implement this and subsequent changes to improve the script, let me know and I will keep these pull requests open until you are confident they can be merged to test.

orazioedoardo:
> If the issue is "don't remove files that the user may have already modified for its own purposes" then it's a mitigation, since if pivpn installed the package those files where not there, so it's safe to remove them.

Ok well there is a case where the user may modify pivpn installed configs after the package is installed.

corbolais:
> If the issue is "don't remove files that the user may have already modified for its own purposes" then it's a mitigation, since if pivpn installed the package those files where not there, so it's safe to remove them.

OK, if this is true for every single file in question.

> > For what it's worth, maybe it should be considered globally implementing the use of `install(1)`.
> 
> If you are willing to implement this and subsequent changes to improve the script, let me know and I will keep these pull requests open until you are confident they can be merged to test.

I can do so. Meanwhile, there's no need to keep this PRs open. They can be merged into test as-is.

Any future improvements go in independently. It will be a rather lengthy process to completely overhaul the scripts, so it will happen over a stretch of time.

corbolais:
> > If the issue is "don't remove files that the user may have already modified for its own purposes" then it's a mitigation, since if pivpn installed the package those files where not there, so it's safe to remove them.
> 
> Ok well there is a case where the user may modify pivpn installed configs after the package is installed.

ACK

orazioedoardo:
> I can do so. Meanwhile, there's no need to keep this PRs open. They can be merged into test as-is.

I will merge https://github.com/pivpn/pivpn/pull/876, https://github.com/pivpn/pivpn/pull/880 and https://github.com/pivpn/pivpn/pull/878. Not https://github.com/pivpn/pivpn/pull/879 as it overlaps with https://github.com/pivpn/pivpn/pull/880.

corbolais:
> > I can do so. Meanwhile, there's no need to keep this PRs open. They can be merged into test as-is.
> 
> I will merge #876, #880 and #878. Not #879 as it overlaps with #880.
Not sure what happened w/ 47c84e6a45a9142800f4669cf4a5e72ac0c36122, it ended up identically in two PRs, of which only #879 was intended.

orazioedoardo:
#880 has all the changes for the uninstall.sh script from #879 plus changes for the install.sh. Are you discarding the latter?

corbolais:
Uh. I will investigate this. Sorry. Lots of PRs on top of each other or depending on one or another.

corbolais:
> #880 has all the changes for the uninstall.sh script from #879 plus changes for the install.sh. Are you discarding the install.sh? (I will fix the merge conflict If you choose #880)

Ok, the way I meant this is: I originally intended commit 47c84e6a45a9142800f4669cf4a5e72ac0c36122 for PR #879, with #879 being the PR for `update.sh`. 

However, additionally, commit 47c84e6a45a9142800f4669cf4a5e72ac0c36122 somehow erroneously ended up in PR #880. 

So, as it is now, it seems best to drop PR #879 and just use PR #880 as commit 47c84e6a45a9142800f4669cf4a5e72ac0c36122 is already part of #880.

Does that make sense to anyone else? ;-)

Changes have been merged.

Thanks.

PR #882, opened 2019-12-10
Revert some minor changes

orazioedoardo:
- Unquoted ${PKG_INSTALL} and ${UPDATE_PKG_CACHE} as they seem to hang the script if quoted.
- Removed [this code](https://github.com/pivpn/pivpn/compare/test...orazioedoardo:test?expand=1#diff-299d9b30c952310aace4aba77c698874L285-L294) as set -e is not used in the test branch and even if it were used, the script would exit if `debconf-apt-progress`'s exit code is NOT 100 (which is the case when no errors occur, exit code 0).
- `vpnGw="${pivpnNET/.0/.1}"` turns 10.8.0.0 into 10.8.1.0 however we need 10.8.0.1 so `vpnGw="${pivpnNET/.0.0/.0.1}"` is used.
- Create folder /var/lib/openvpn
- Removed code path related to OLD_UFW, originally used for jessie compatibility
- Other trivial changes
@corbolais what do you think?

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1432,10 +1432,10 @@ confNetwork(){
 			fi
 			### If there is already a "*nat" section just add our POSTROUTING MASQUERADE
 			if $SUDO grep -q "*nat" /etc/ufw/before.rules; then
-				$SUDO sed "/^*nat/{n;s/\(:POSTROUTING ACCEPT .*\)/\1\n-I POSTROUTING -s ${pivpnNET}\/24 -o ${IPv4dev} -j MASQUERADE/}" -i /etc/ufw/before.rules
+				$SUDO sed "/^*nat/{n;s/\(:POSTROUTING ACCEPT .*\)/\1\n-I POSTROUTING -s ${pivpnNET}\/${subnetClass} -o ${IPv4dev} -j MASQUERADE/}" -i /etc/ufw/before.rules
 			else
-        $SUDO sed "/delete these required/i *nat\n:POSTROUTING ACCEPT [0:0]\n-I POSTROUTING -s ${pivpnNET}\/${subnetClass} -o ${IPv4dev} -j MASQUERADE\nCOMMIT\n" -i /etc/ufw/before.rules
-      fi
+				$SUDO sed "/delete these required/i *nat\n:POSTROUTING ACCEPT [0:0]\n-I POSTROUTING -s ${pivpnNET}\/${subnetClass} -o ${IPv4dev} -j MASQUERADE\nCOMMIT\n" -i /etc/ufw/before.rules
+			fi
 			# Insert rules at the beginning of the chain (in case there are other rules that may drop the traffic)
 			$SUDO ufw insert 1 allow "${pivpnPORT}"/"${pivpnPROTO}" >/dev/null
 			$SUDO ufw route insert 1 allow in on "${pivpnDEV}" from "${pivpnNET}/${subnetClass}" out on "${IPv4dev}" to any >/dev/null
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -19,6 +19,7 @@ debianOvpnUserGroup="openvpn:openvpn"
 ### PKG Vars ###
 PKG_MANAGER="apt-get"
 PKG_CACHE="/var/lib/apt/lists/"
+### FIXME: quoting UPDATE_PKG_CACHE and PKG_INSTALL hangs the script, shellcheck SC2086
 UPDATE_PKG_CACHE="${PKG_MANAGER} update"
 PKG_INSTALL="${PKG_MANAGER} --yes --no-install-recommends install"
 PKG_COUNT="${PKG_MANAGER} -s -o Debug::NoLocking=true upgrade | grep -c ^Inst || true"
@@ -1308,8 +1309,7 @@ set_var EASYRSA_KEY_SIZE   ${pivpnENCRYPT}" | $SUDO tee vars >/dev/null
 	${SUDOE} ./easyrsa gen-crl
 	${SUDOE} cp pki/crl.pem /etc/openvpn/crl.pem
   if ! getent passwd openvpn; then
-    mkdir -p /var/lib/openvpn
-    ${SUDOE} adduser --system --home /var/lib/openvpn/ --no-create-home --group --disabled-login ${debianOvpnUserGroup%:*}
+    ${SUDOE} adduser --system --home /var/lib/openvpn/ --group --disabled-login ${debianOvpnUserGroup%:*}
   fi
   ${SUDOE} chown "$debianOvpnUserGroup" /etc/openvpn/crl.pem
 
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -227,7 +227,7 @@ updatePackageCache(){
 		#update package lists
 		echo ":::"
 		echo -ne "::: ${PKG_MANAGER} update has not been run today. Running now...\\n"
-		$SUDO "${UPDATE_PKG_CACHE}" &> /dev/null
+		$SUDO ${UPDATE_PKG_CACHE} &> /dev/null
 		echo " done!"
 	fi
 }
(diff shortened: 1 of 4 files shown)


corbolais:
- regarding `set -e` in test: good catch

> the script would exit if debconf-apt-progress's exit code is NOT 100 (which is the case when no errors occur, exit code 0)

- true. ACK. 
Intention was: do NOT exit on return code 100. I got reproducibly ERR 100 but installation and configuration went fine. So, when installing on bullseye/sid the installation of pivpn will simply drop out. That's why for this distribution version a condition is required.

> vpnGw="${pivpnNET/.0/.1}" turns 10.8.0.0 into 10.8.1.0 however we need 10.8.0.1 so vpnGw="${pivpnNET/.0.0/.0.1}" is used.

- That is strange. I must have tested this but inserted something else. You are right, of course.

> Create folder /var/lib/openvpn

- Good idea. But just omit `--no-create-home` rather than use mkdir.

> Unquoted ${PKG_INSTALL} and ${UPDATE_PKG_CACHE} as they seem to hang the script if quoted.

- Just merge your change. I think this was a shellcheck issue. I'll have a look at it at some point. Would you mind adding a FIXME for me as a reminder at the top of the file?

- `scripts/openvpn/pivpnDebug.sh`: use pivpnNET/subnetClass instead of IP/24

- I'm not sure how this got left out: https://github.com/pivpn/pivpn/pull/882/commits/c20e7d21502044c2dc80e80504405c98e78668b5#diff-299d9b30c952310aace4aba77c698874L1435 . Good catch.

- `server_config.txt`: Good catch, again.

- `scripts/uninstall.sh`: `rm -rf` Are those really directories? Also, I'm wondering whether it's a good idea to remove old configs or certs (ovpns). Maybe the user needs those but wants to just remove the service or cleanly reinstall it? Touching user's stuff is touchy..

Thank you.

orazioedoardo:
> Intention was: do NOT exit on return code 100. I got reproducibly ERR 100 but installation and configuration went fine. So, when installing on bullseye/sid the installation of pivpn will simply drop out. That's why for this distribution version a condition is required.

Without `set -e` (default case) it should not exit.

> scripts/openvpn/pivpnDebug.sh: use pivpnNET/subnetClass instead of IP/24

I changed uninstall.sh to get started. Will do all other scripts that have /24 hardcoded later. `subnetClass`, `pivpnDEV`, `pivpnNET`, `vpnGw` could be sourced from `setupVars.conf`

> scripts/uninstall.sh: rm -rf Are those really directories? Also, I'm wondering whether it's a good idea to remove old configs or certs (ovpns). Maybe the user needs those but wants to just remove the service or cleanly reinstall it? Touching user's stuff is touchy..

https://github.com/pivpn/pivpn/blob/ca11f796293fed333b984362325fb9e16742cd7b/auto_install/install.sh#L1401

https://github.com/pivpn/pivpn/blob/ca11f796293fed333b984362325fb9e16742cd7b/auto_install/install.sh#L1403

https://github.com/pivpn/pivpn/blob/ca11f796293fed333b984362325fb9e16742cd7b/auto_install/install.sh#L1268

If we delete those three directories it makes little sense to retains the client certs and configs, they can't be used anymore.

corbolais:
> > Intention was: do NOT exit on return code 100. I got reproducibly ERR 100 but installation and configuration went fine. So, when installing on bullseye/sid the installation of pivpn will simply drop out. That's why for this distribution version a condition is required.
> 
> Without `set -e` (default case) it should not exit.

Well then, OK.

> 
> > scripts/openvpn/pivpnDebug.sh: use pivpnNET/subnetClass instead of IP/24
> 
> I changed uninstall.sh to get started. Will do all other scripts that have /24 hardcoded later. `subnetClass`, `pivpnDEV`, `pivpnNET`, `vpnGw` could be sourced from `setupVars.conf`
> 
> > scripts/uninstall.sh: rm -rf Are those really directories? Also, I'm wondering whether it's a good idea to remove old configs or certs (ovpns). Maybe the user needs those but wants to just remove the service or cleanly reinstall it? Touching user's stuff is touchy..
> 
> https://github.com/pivpn/pivpn/blob/ca11f796293fed333b984362325fb9e16742cd7b/auto_install/install.sh#L1401
> 
> https://github.com/pivpn/pivpn/blob/ca11f796293fed333b984362325fb9e16742cd7b/auto_install/install.sh#L1403
> 
> https://github.com/pivpn/pivpn/blob/ca11f796293fed333b984362325fb9e16742cd7b/auto_install/install.sh#L1268
> 
> If we delete those three directories as well as `openvpn.conf` and `wg0.conf` it makes little sense to retain the client certs and configs, they can't be used anymore.
> 

So, keep them alltogether. Most likely, we should ask what the user wants us to with them, anyway: Keep || do them away. This is true for any such stuff as `rm stuff`, fw rules included. But most importantly, discarding user stuff that cannot be restored or recreated is a big nono, IMO. If unsure, let the user clean up on it's own. That way nothing is lost, on our behalves, at the least.

corbolais:
> I changed uninstall.sh to get started. Will do all other scripts that have /24 hardcoded later. `subnetClass`, `pivpnDEV`, `pivpnNET`, `vpnGw` could be sourced from `setupVars.conf`

Perfect.

`vpnGw` would be a candidate for `setupVars.conf` only if it is customizable and actually implemented in the scripts that way, wouldn't it? Otherwise we'd just derive it from `pivpnNET`.

Ideally, pivpn would verify identity between settings in `setupVars.conf` and the actual system configuration, i.e. interfaces, network, netmask, iptables rules etc. That way, we wouldn't be scrapping the users' manual modifications, potentially rendering systems, networks and businesses inoperable. Verify and yell the discrepancies at the user for them to be checked and resolved.

orazioedoardo:
> So, keep them alltogether. Most likely, we should ask what the user wants us to with them, anyway: Keep || do them away.

I mean, take the openvpn server. If we remove the easy-rsa folder, which contains the public key infrastructure, the .ovpn files simply can’t be used because the certificates inside refer to the cerificates in the pki. Same with wireguard. If you delete the server config, which contains public/private keys, then client configs are useless. So why keep them anyways?

> This is true for any such stuff as rm stuff, fw rules included.

In my opinion, the uninstall script should undo all changes to the system the install script did. What do you think an uninstall script should do?

> vpnGw would be a candidate for setupVars.conf only if it is customizable and actually implemented in the scripts that way, wouldn't it? Otherwise we'd just derive it from pivpnNET.

Apart from subnetClass, all of the above variables can just be derived from the VPN variable. However if we decide not to hardcode CIDRs inside the script it makes sense to me to derive them once and source them in all script that use them (look at the openvpn and wireguard folder).

> Ideally, pivpn would verify identity between settings in setupVars.conf and the actual system configuration, i.e. interfaces, network, netmask, iptables rules etc.

Good idea

corbolais:
> > So, keep them alltogether. Most likely, we should ask what the user wants us to with them, anyway: Keep || do them away.
> 
> I mean, take the openvpn server. If we remove the easy-rsa folder, which contains the public key infrastructure, the .ovpn files simply can’t be used because the certificates inside refer to the cerificates in the pki. Same with wireguard. If you delete the server config, which contains public/private keys, then client configs are useless. So why keep them anyways?

I mean, at least ask and double-check whether the user really intents us to unlink her certs and stuff. I personally prefer having the "old" certs around in a versioned/dated backup-directory for some time longer. It happens that someone only realises the need after days/weeks passed by.

> > This is true for any such stuff as rm stuff, fw rules included.
> 
> In my opinion, the uninstall script should undo all changes to the system the install script did. What do you think an uninstall script should do?

:-D
My point is, we should verify whether what we rm is definitely our stuff and nothing handled by the user. I'd prefer having stuff left around in case. Nothing worse than rm'ing stuff only to realize, wait a minute, where did xy end up!?

Take for example a fw rule: we should verify user doesn't really need it any longer. Ask. Verify it's really the rule we installed etc. Same with files. If in doubt, leave the cruft and tell the user we leave it on purpose because it seems like it's something she might need (maybe).
 
> > vpnGw would be a candidate for setupVars.conf only if it is customizable and actually implemented in the scripts that way, wouldn't it? Otherwise we'd just derive it from pivpnNET.
> 
> Apart from subnetClass, all of the above variables can just be derived from the VPN variable. However if we decide not to hardcode CIDRs inside the script it makes sense to me to derive them once and source them in all script that use them (look at the openvpn and wireguard folder).

Yes, that's a way. If we verify stuff anyway, we can dump all config in `setupVars.conf`. Sure thing.

orazioedoardo:
> I mean, at least ask and double-check whether the user really intents us to unlink her certs and stuff. I personally prefer having the "old" certs around in a versioned/dated backup-directory for some time longer. It happens that someone only realises the need after days/weeks passed by.

OK

> :-D
My point is, we should verify whether what we rm is definitely our stuff and nothing handled by the user. I'd prefer having stuff left around in case. Nothing worse than rm'ing stuff only to realize, wait a minute, where did xy end up!?

Take for example a fw rule: we should verify user doesn't really need it any longer. Ask. Verify it's really the rule we installed etc. Same with files. If in doubt, leave the cruft and tell the user we leave it on purpose because it seems like it's something she might need (maybe).

Yes delete stuff but make sure that it's actually what pivpn added. We could "watermark" files and configs (for example specific file name, disclaimer inside config files, iptables rule comment).

corbolais:
> Yes delete stuff but make sure that it's actually what pivpn added. We could "watermark" files and configs (for example specific file name, disclaimer inside config files, iptables rule comment).

That's a good point!

PR #891, opened 2019-12-24
Supersede Google DNS for privacy

h-town:
Removed & superseded default reference to Google DNS services with the addresses for Quad9.  Both Google and Cloudflare were retained as options in these edits but I would encourage an additional notice to users for both providers.

I am not affiliated with Quad9.  It was not included by default and thought I'd add while removing.


auto_install/install.sh
```diff
@@ -947,8 +947,8 @@ askClientDNS(){
 			pivpnDNS1="$pivpnDNS2"
 			unset pivpnDNS2
 		elif [ -z "$pivpnDNS1" ] && [ -z "$pivpnDNS2" ]; then
-			pivpnDNS1="8.8.8.8"
-			pivpnDNS2="8.8.4.4"
+			pivpnDNS1="9.9.9.9"
+			pivpnDNS2="149.112.112.112"
 			echo "::: No DNS provider specified, using Google DNS ($pivpnDNS1 $pivpnDNS2)"
```
orazioedoardo: You should update this line too.

auto_install/install.sh
```diff
@@ -57,8 +57,8 @@ r=$(( r < 20 ? 20 : r ))
 c=$(( c < 70 ? 70 : c ))
 
 # Find IP used to route to outside world
-IPv4addr=$(ip route get 8.8.8.8| awk '{print $7}')
-IPv4gw=$(ip route get 8.8.8.8 | awk '{print $3}')
+IPv4addr=$(ip route get 9.9.9.9 | awk '{print $7}')
+IPv4gw=$(ip route get 9.9.9.9 | awk '{print $3}')
```
orazioedoardo: This can be any public IP address since it doesn't actually contact the server. While we are here can you replace the references with 192.0.2.1?

From [RFC 5735](https://tools.ietf.org/html/rfc5735):
> 192.0.2.0/24 - This block is assigned as "TEST-NET-1" for use in documentation and example code.  It is often used in conjunction with domain names example.com or example.net in vendor and protocol documentation.  As described in [RFC5737], addresses within this block do not legitimately appear on the public Internet and can be used without any coordination with IANA or an Internet registry.  See [RFC1166].

server_config.txt
```diff
@@ -8,8 +8,8 @@ dh /etc/openvpn/easy-rsa/pki/dh2048.pem
 topology subnet
 server 10.8.0.0 255.255.255.0
 # Set your primary domain name server address for clients
-push "dhcp-option DNS 8.8.8.8"
-push "dhcp-option DNS 8.8.4.4"
+push "dhcp-option DNS 9.9.9.9"
+push "dhcp-option DNS 149.112.112.112"
```
orazioedoardo: Same as [here](https://github.com/pivpn/pivpn/pull/891/files#r362763211). These IPs will be replaced upon installation.

auto_install/install.sh
h-town: Thanks @orazioedoardo! I'm not sure why these didn't show up in my search.

server_config.txt
h-town: I liked the original approach of using "actual" resolvers as a just-in-case; I just thought I'd change the just-in-case to the suggested default.  Best to leave as-is so as not to cause confusion from those reviewing the script before executing?

orazioedoardo: Good point.

PR #932, opened 2020-01-31
Test to master

4s3ti:
* added link to server status dashboard

* Replaced Header with bold instead

* More safeguards, some fixes, standardized some code, WireGuard update script, removed redundant code

  - Add curl as a dependency for those who run the script without 'curl URL | bash'.
  - Use POSIX 'command -v' instead of 'hash'.
  - Check if packages have actually been installed and abort execution if they have not.
  - Fixed issue with getStaticIPv4Settings() that prevented existing network settings
    to be used as static IP settings when running the script unattended with empty
    $IPv4addr and $IPv4gw variables.
  - Exit if processing wireguard-linux-compat fails.
  - Exit if 50unattended-upgrades fails to extract.
  - Exit clientSTAT.sh if the wg0 interface is not available.
  - Moved the Self Check to a single script since dedicated versions were very similar.
  - Add 'pivpn -wg' to update WireGuard for users running Raspbian with armv6l kernel.

* Fixed cosmetic issue with spinner, added missing spinner to some APT commands

* Detect current netmask, validate user input when configuring a static IP

* Inform the user when updating the package cache, which can be slow on some RPis

* Invalidate $IPv4Addr and $IPv4gw when the user claims those settings are not correct

* Restart pihole in the more appropriate restartServices() function

* Improve static IP selection, validate public DNS name of the server
  - Default to 'No' when asking if the RPi has DHCP reservation, considered
    that the user may not be fully aware, furthermore, setting a static IP
    anyways doesn't do harm.
  - Validate existing IPv4 settings (address, gateway, DNS) to avoid filling
    '/etc/dhcpcd.conf' with invalid data.
  - Validate public DNS name of the server inside askPublicIPOrDNS() function

* Check DH parameters, fix 'pivpn -c', improvements when dealing with external repositories
  - Added a basic sanity check to downloaded DH paramenters, which doubles as a
    check for missing .pem file.
  - Fix 'pivpn -c' showing the month number instead of the day of the month when
    using WireGuard.
  - Removing APT keys is risky, it would break APT update/upgrade if the user
    already was already using the unstable repo.
  - Replaced 'Checking for $i... installed' in favor of a more clear 'Checking for
    $i... already installed'.
  - Check whether the OpenVPN repo and the Debian unstable repo are already used.

* Improvements to getStaticIPv4Settings()

  - Use a regular expression to extract IPs from the 'ip' command. With this,
    there is a little need to validate output. Even though the regex will match
    invalid IPs like 192.168.23.444, 'ip' can't return them, and even if it did,
    the script would not have reached this function due to previous functions
    using the network with broken routes and addresses.

  - Get the IP address from the selected interface rather then from the 'ip route'
    command as it's not guaranteed that such IP is the same of the interface the
    user decided to use (though on a Raspberry Pi inside a home LAN, most likely
    it is, but it also maskes easier to get the IP in the CIDR notation with a
    single 'ip | grep' pipe).

* Moved command substitution to specific functions to avoid unnecessary execution

  - Moved $availableInterfaces and $CurrentIPv4gw from the script header to
    their relevant function, considered that if the OS is not Raspbian a static
    IP is not set, so those variables are not used.

* Copy files from git repo using the 'install' command, switch DH params from 2ton.com.au to RFC 7919

  - Now using DH parameters suggested by the RFC 7919 for use by TLS servers (the user can
    still generate his own if he wishes).
    https://wiki.mozilla.org/Security/Archive/Server_Side_TLS_4.0#Pre-defined_DHE_groups


4s3ti:
@orazioedoardo here .. check if all looks as it should and let me know ... 

then we just rebase test. 
was kinda my bad here.. as i don't really like to squash commits.

PR #950, opened 2020-02-15
Merge test into master

orazioedoardo:
Main changes from the test branch:
1) When offering to use Pi-hole, identify VPN clients via clientname.pivpn using a dedicated hosts file. Clients can now be resolved by their names and also show up in the Pi-hole dashboard.
2) Decide whether to tell dnsmasq to listen on the VPN interface depending on the user settings. The default Pi-hole listening behavior is "Listen only on interface whatever", which means we need to add the specific VPN interface. However, if the user has  "Listen on all interfaces" or " "Listen on all interfaces, permit all origin", then we don't need to add the interface (self-explanatory).
3) Set static IPs by default when using OpenVPN, (required by 2.).
4) Restrict access to automatic backups (.tar.gz) of /etc/wireguard and /etc/openvpn to root.

### DIFF
--- a/scripts/openvpn/makeOVPN.sh
+++ b/scripts/openvpn/makeOVPN.sh
@@ -405,6 +405,17 @@ else
 
 fi
 
+# Find an unused number for the last octet of the client IP
+for i in {2..254}; do
+    # find returns 0 if the folder is empty, so we create the 'ls -A [...]'
+    # exception to stop at the first static IP (10.8.0.2). Otherwise it would
+    # cycle to the end without finding and available octet.
+    if [ -z "$(ls -A /etc/openvpn/ccd)" ] || ! find /etc/openvpn/ccd -type f -exec grep -q "10.8.0.$i" {} +; then
+        COUNT="$i"
+        echo "ifconfig-push 10.8.0.$i 255.255.255.0" >> /etc/openvpn/ccd/"${NAME}"
+        break
+    fi
+done
 
 # Copy the .ovpn profile to the home directory for convenient remote access
 cp "/etc/openvpn/easy-rsa/pki/$NAME$FILEEXT" "$install_home/ovpns/$NAME$FILEEXT"
--- a/scripts/openvpn/makeOVPN.sh
+++ b/scripts/openvpn/makeOVPN.sh
@@ -405,20 +405,29 @@ else
 
 fi
 
+cidrToMask(){
+	# Source: https://stackoverflow.com/a/20767392
+	set -- $(( 5 - ($1 / 8) )) 255 255 255 255 $(( (255 << (8 - ($1 % 8))) & 255 )) 0 0 0
+	[ $1 -gt 1 ] && shift $1 || shift
+	echo ${1-0}.${2-0}.${3-0}.${4-0}
+}
+
+NET_REDUCED="${pivpnNET::-2}"
+
 # Find an unused number for the last octet of the client IP
 for i in {2..254}; do
     # find returns 0 if the folder is empty, so we create the 'ls -A [...]'
     # exception to stop at the first static IP (10.8.0.2). Otherwise it would
     # cycle to the end without finding and available octet.
-    if [ -z "$(ls -A /etc/openvpn/ccd)" ] || ! find /etc/openvpn/ccd -type f -exec grep -q "10.8.0.$i" {} +; then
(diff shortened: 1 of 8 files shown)


scripts/openvpn/makeOVPN.sh
```diff
@@ -405,6 +405,26 @@ else
 
 fi
 
+# Find an unused number for the last octet of the client IP
+for i in {2..254}; do
+    # find returns 0 if the folder is empty, so we create the 'ls -A [...]'
+    # exception to stop at the first static IP (10.8.0.2). Otherwise it would
+    # cycle to the end without finding and available octet.
+    if [ -z "$(ls -A /etc/openvpn/ccd)" ] || ! find /etc/openvpn/ccd -type f -exec grep -q "10.8.0.$i" {} +; then
```
4s3ti: Greping for 10.8.0.$i doesn't seem right .. there might be cases where users change the vpn IP ranges ... instead should  pickup from the setup vars and then do the rest of the magic

scripts/openvpn/makeOVPN.sh
```diff
@@ -405,6 +405,26 @@ else
...
+        COUNT="$i"
+        echo "ifconfig-push 10.8.0.$i 255.255.255.0" >> /etc/openvpn/ccd/"${NAME}"
```
4s3ti: same here

scripts/openvpn/makeOVPN.sh
```diff
@@ -405,6 +405,26 @@ else
...
+        break
+    fi
+done
+
+if [ -f /etc/pivpn/hosts.openvpn ]; then
+    echo "10.8.0.${COUNT} ${NAME}.pivpn" >> /etc/pivpn/hosts.openvpn
```
4s3ti: and here

scripts/openvpn/removeOVPN.sh
```diff
@@ -120,8 +120,22 @@ for (( ii = 0; ii < ${#CERTS_TO_REVOKE[@]}; ii++)); do
     rm -rf "pki/private/${CERTS_TO_REVOKE[ii]}.key"
     rm -rf "pki/issued/${CERTS_TO_REVOKE[ii]}.crt"
 
+    # Grab the client IP address
+    STATIC_IP=$(grep -v "^#" /etc/openvpn/ccd/"${CERTS_TO_REVOKE[ii]}" | grep -w ifconfig-push | grep -oE '10.8.0\.[0-9]{1,3}')
```
4s3ti: same situation here with the 10.8

scripts/wireguard/makeCONF.sh
```diff
@@ -113,6 +113,15 @@ AllowedIPs = 10.6.0.${COUNT}/32
 # end ${CLIENT_NAME}" >> wg0.conf
 echo "::: Updated server config"
 
+if [ -f /etc/pivpn/hosts.wireguard ]; then
+    echo "10.6.0.${COUNT} ${CLIENT_NAME}.pivpn" >> /etc/pivpn/hosts.wireguard
```
4s3ti: And here.. 10.6

scripts/wireguard/removeCONF.sh
```diff
@@ -101,9 +101,19 @@ for CLIENT_NAME in "${CLIENTS_TO_REMOVE[@]}"; do
                 fi
             done
 
+            ((DELETED_COUNT++))
             echo "::: Successfully deleted ${CLIENT_NAME}"
 
-            ((DELETED_COUNT++))
+            # If using Pi-hole, remove the client from the hosts file
+            if [ -f /etc/pivpn/hosts.wireguard ]; then
+                sed "\#10.6.0.${COUNT} ${CLIENT_NAME}.pivpn#d" -i /etc/pivpn/hosts.wireguard
```
4s3ti: Also here .. sed 10.6  ...

orazioedoardo:
Done, now VPN ranges are defined solely by the variables `pivpnNET` and `subnetClass`.

orazioedoardo:
> Generally looks good.. i assume you have been testing this aigh? xD i din't had much time lately for testing

Just wanted to know it you liked the commit that removed hard coded VPN ranges. I did test these on Ubuntu 18.04, Debian 10 and Raspbian 10.

Issue #968, opened 2020-03-01
[FR] install both Wireguard and Openvpn


sckemu:
**Is your feature request related to a problem? Please describe.**
I usually want to use Wireguard, but I want to use OpenVPN on networks where UDP doesn't work.

**Describe the solution you'd like**
want to install both Wireguard and Openvpn.
Add client:
```
pivpn add ovpn
pivpn add wg
```

shelleycat485:
I wanted this too, and have got it working, but very manual.  
1 copy installation scripts to pivpnx and pivpny installers.
2  Edited both so that they would install into /optpivpnx and /opt/pivpny.  Changed repository locations in /etc/.pivpnX and Y so I will have two repositories at end.
3  Installed twice, once from each of the scripts.  Once ovpn, once wireguard.  
4  Then had to edit /etc/bash_completion.d (two entries needed)  
5  had to edit the /usr/local/bin/pivpnX and pivpnY to point to the right /opt locations
6  had to edit many of the scripts in the /opt locations, edits are changing locations from pivpn to pivpnX or Y were they are locations (and not part of a variable)
7 update scripts is not going to work.  Most other things now seem to, I've tested both protocols now.  So openvpn is pivpnx and wireguard is pivpny,  On /etc/debian-version 9.11
I will be looking at whether I can split out the initial location somewhere, and then substitute into the right places in the scripts, but there are a lot of edits to do for that, to try and avoid steps 4,5,6 and get 7 working.  Maybe put the /opt location into the setupvars?

shelleycat485:
One solution in https://github.com/shelleycat485/pivpn


Added --altcommand to the installer.
Allows two installations at once.
The installations can be removed or updated independently.
Example bash install.sh --altcommand wg
  then try 'pivpn_wg help'
Example bash install.sh --altcommand opv
  then try 'pivpn_opv help'
or if already installed, can install the other VPN protocol as well
Example bash install.sh
Example bash install.sh --altcommand opv
The --altcommand names are not checked against the vpn type

constructor-s:
I know this is not what is originally asked for, but I simply installed Wireguard with PiVPN and then OpenVPN with another script like https://github.com/angristan/openvpn-install

shelleycat485:
I have got both working, but currently with two different commands.  Also the update and uninstall are both going to cause problems if I ever use them.
[...]

PR #1029, opened 2020-04-23
 Install both Wireguard and Openvpn protocols (Issue #968)

shelleycat485:
Added --altcommand to the installer.
        Allows two installations at once.
        The installations can be removed or updated independently.
        Example bash install.sh --altcommand wg
          then try 'pivpn_wg help'
        Example bash install.sh --altcommand opv
          then try 'pivpn_opv help'
        or if already installed, can install the other VPN protocol as well
        Example bash install.sh
        Example bash install.sh --altcommand opv
        The --altcommand names are not checked against the vpn type

Tested using RasPIW+.  Because not a major use case this has changed the command  for at least one of the protocols.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -12,8 +12,10 @@
 
 ######## VARIABLES #########
 pivpnGitUrl="https://github.com/pivpn/pivpn.git"
-setupVars="/etc/pivpn/setupVars.conf"
-pivpnFilesDir="/etc/.pivpn"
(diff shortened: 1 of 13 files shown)


4s3ti:
Hi Thank you for your MR, however it doesn't follow our contributor guidelines. 

Please read: https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md

Pull requests should be done to test branch and not master. 

Thank you.

shelleycat485:
Sorry, missed that.  Will try again.

PR #1030, opened 2020-04-23
Install both Wireguard and Openvpn protocols (Issue #968) 

shelleycat485:
Added --altcommand to the installer.
    Allows two installations at once.
    The installations can be removed or updated independently.
    Example bash install.sh --altcommand wg
      then try 'pivpn_wg help'
    Example bash install.sh --altcommand opv
      then try 'pivpn_opv help'
    or if already installed, can install the other VPN protocol as well
    Example bash install.sh
    Example bash install.sh --altcommand opv
    The --altcommand names are not checked against the vpn type

Tested using RasPiW+. Because not a major use case this has changed the command for at least one of the protocols.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -12,8 +12,10 @@
 
 ######## VARIABLES #########
 pivpnGitUrl="https://github.com/pivpn/pivpn.git"
-setupVars="/etc/pivpn/setupVars.conf"
-pivpnFilesDir="/etc/.pivpn"
+setupVars="/etc/pivpn/setupVars.conf" #can be modified if --altcommand
+pivpnetcFilesDir="/etc/.pivpn"        #can be modified if --altcommand
+pivpnoptFilesDir="/opt/pivpn"         #can be modified if --altcommand
+pivpnlocalbinFilesDir="/usr/local/bin/pivpn" #can be modified if --altcommand
 
 piholeSetupVars="/etc/pihole/setupVars.conf"
 dnsmasqConfig="/etc/dnsmasq.d/02-pivpn.conf"
@@ -56,6 +58,7 @@ OPENVPN_KEY_URL="https://swupdate.openvpn.net/repos/repo-public.gpg"
 runUnattended=false
 skipSpaceCheck=false
 reconfigure=false
+altcommandname=""
 
 ######## SCRIPT ########
 
@@ -95,11 +98,15 @@ main(){
 	fi
 
 	# Check arguments for the undocumented flags
-	for var in "$@"; do
-		case "$var" in
+        args=$#
+	for (( i=0; i<=$args; i+=1 ))
+        do
+                argarray=("$@")
+		case "${argarray[$i]}" in
 			"--i_do_not_follow_recommendations"   ) skipSpaceCheck=false;;
 			"--unattended"     ) runUnattended=true;;
 			"--reconfigure"  ) reconfigure=true;;
+                        "--altcommand"   ) altcommandcheck "${argarray[$i+1]}";;
 		esac
 	done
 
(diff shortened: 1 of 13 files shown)


orazioedoardo:
I was expecting that someone asked for this, and me too. However my idea was a bit different: unify the [pivpn script](https://github.com/pivpn/pivpn/issues/968#issue-573561287) as well as bash_completion. Place script inside `/opt/pivpn/$VPN` (or just copy [this folder](https://github.com/pivpn/pivpn/tree/master/scripts) from the git repo to `/opt/pivpn`) and setup vars file inside `/etc/pivpn/$VPN/setupVars.conf`. Then, when the user runs the script again over an existing installation (`askAboutExistingInstall()`) also provide an option to install the other VPN protocol, which should not overlap considering the changes in the folder structure above.

shelleycat485:
I can follow your idea on that if you want.  I chose the dual structure because I had a pi working at a remote site, and was scared of breaking my ovpn vpn into the remote site (lockdown!).  So I tried to make the two installations as separate as possible, so either protocol could be updated/patched and the other one is still working.

orazioedoardo:
I can follow your idea on that if you want.  I chose the dual structure because I had a pi working at a remote site, and was scared of breaking my ovpn vpn into the remote site (lockdown!).

> That’s dangerous, for most tests I usually run a Debian virtual machine unless the change is Raspberry Pi specific. In this case a virtual machine should be fine.

shelleycat485:
I tested what I was doing on a pi at home, but I wanted a solution I was sure would not break a remote machine.

shelleycat485:
I'm looking at moving/unifying the pivpn scripts.  Should I just leave this here until another solution comes along, or do I delete it, or does someone else delete it?

tlkcode:
What is the best way to test this? I cloned the repo, and do the successful install with the --altcommand, but in the end, the pivpn_wg script has issues because it is looking for the scripts in /opt/pivpn rather than /opt/pivpn_wg . Updating them, all of the other scripts seem to need updates too, I think because of the makeRepo git to /etc/.pivpn.

shelleycat485:
Hi, assuming you cloned the ...shelleycat485 repository, did you clone the right branch? I think the branch you want is test.  Then on your local machine copy out the install.sh and do a bash install.sh  This is different from the normal curl... install because of the different branch.  I'll try and verify that as well.  All the changes should have been in step, and it sounds like you are accessing the wrong branch.

orazioedoardo:
> I'm looking at moving/unifying the pivpn scripts. Should I just leave this here until another solution comes along, or do I delete it, or does someone else delete it?

Whatever you want, close the pull request and open a new one when you have the changes ready or keep this open and add new commits

tlkcode:
Yes, I got the right branch (git clone -b test https://github.com/shelleycat485/pivpn/), and all of the scripts have the dual installation capability in the repo, but when I install with the --altcommand, they get the pivpn/pivpn version. I think in install.sh, the install clones down the pivpn/pivpn version and uses that. After install, I edited /usr/local/bin/pivpn_opv to reference the correct path, and copied the OpenVPN scripts in shelleycat485:test cloned repo the to /opt/pivpn_opv folders. Things almost seem to work now, but I think there is still some cross contamination where my OpenVPN clients are getting assigned ip ranges out of the WG pool.

shelleycat485:
OK, thanks for that testing information.  I had not noticed the cross-contamination of the addresses.  I'll have  a look at that - I tested it with one client of each sort, so I will try more clients
I understand the install.sh issue, I think, the ...shelleycat repo is set up to run from pivpn/pivpn on github because it is a pull request, but if you are using it locally line 14 needs to point at the ...shelleycat.. repo on github.

tlkcode:
I can confirm no cross contamination when installed from the local repo and the install.sh is edited to use shelleycat test. Both WG & OpenVPN tested and working.

shelleycat485:
Will close this, working on a solution that will retains pivpn as the single command.  I'll leave this repo as it is, put a comment to it in the request for dual protocol in the pivpn/pivpn issues to people can find it.

aaomidi:
Did this ever land?

shelleycat485:
Yes, its in the main pivpn/pivpn repository now.

aaomidi:
@shelleycat485 Maybe I'm missing something, but I'm not finding how to install both? I have an OpenVPN setup right now and want to add Wireguard to it. Reinstalling the installation script doesn't give me the option to keep both.

Issue #1038, opened 2020-05-01
[Support] Install on LinuxMint 19.3


RileyKA:
## In raising this issue, I confirm the following: 

`{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting can be *is* directly related to the pivpn installer script.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pivpn/pivpn/wiki/FAQ), [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen)).

### Have you searched for similar issues and solutions?
    (yes/no / which issues?)
yes

### Console output of      `curl -L install.pivpn.dev | bash`
```
  Invalid OS ├───────────────────────────┐
     │                                                                    │ 
     │ We have not been able to detect a supported OS.                    │ 
     │ Currently this installer supports Raspbian, Debian and Ubuntu.     │ 
     │ For more details, check our documentation at                       │ 
     │ https://github.com/pivpn/pivpn/wiki
```

### Console output of      `pivpn add` or `pivpn add nopass`
```
  Output Here
```

### Console output of      `pivpn debug`
```
  Output Here
```
### Have you taken any steps towards solving your issue?
```
  which?
```

getwisp:
Please read the "Prerequisites" section in README.md.
Also, please read the output of the script.
Also, please understand what a feature request is.

If you feel so inclined, I'm sure you are very welcome to find out which packages are required in Linux Mint.

4s3ti:
answer here: https://github.com/pivpn/pivpn/pull/1040

PR #1040, opened 2020-05-04
Added support for Linux Mint 19.3

troygeiger:
I wanted to install pivpn onto my Linux Mint installation. Here are the changes I've made to get it to install. I have only tried with the OpenVPN option and haven't tried to uninstall so I cannot confirm that those areas work. I would think so since Linux Mint is based on Ubuntu.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -258,7 +258,7 @@ distroCheck(){
 		OSCN=${VER_MAP["${VER}"]}
 	fi
 
-	if [ "$PLAT" = "Debian" ] || [ "$PLAT" = "Ubuntu" ]; then
+	if [ "$PLAT" = "Debian" ] || [ "$PLAT" = "Ubuntu" ] || [ "$PLAT" = "LinuxMint" ]; then
 		DPKG_ARCH="$(dpkg --print-architecture)"
 		if [ "$DPKG_ARCH" = "amd64" ] || [ "$DPKG_ARCH" = "i386" ]; then
 			X86_SYSTEM=1
@@ -268,9 +268,9 @@ distroCheck(){
 	fi
 
 	case ${PLAT} in
-		Debian|Raspbian|Ubuntu)
+		Debian|Raspbian|Ubuntu|LinuxMint)
 			case ${OSCN} in
-				buster|xenial|bionic|stretch)
+				buster|xenial|bionic|stretch|tricia)
 				:
 				;;
 				*)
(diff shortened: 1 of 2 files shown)


troygeiger:
Unfortunately not. I hope that they will add support for Pi someday. I know pivpn wasn't ment to be for desktop but it seems to work fine. My Pi died a couple of days ago and I wanted to get my VPN back up and pivpn makes it way easier to setup. I see someone else is also requesting Mint support on issue #1038

4s3ti:
Sorry ... not accepting this one, for multiple reasons ...

1 - lack of testing ... 
2 - Mint is based on debian so it should give you a warning but it should work anyway. 
3 - We cannot keep up with the releases of every distribution ... so we are sticking with 2 main ones .. ubuntu and debian for now. 
4 - Mint doesn't run on a Pi, support for ubuntu, reason why ubuntu is supported is because it runs also in most inexpensive VPS thus allowing those who want a VPN offshore to have it using pivpn. 
5 - Mint is a desktop distribution therefore not the right place to run a VPN server and would be against our posture of encouraging security and privacy best practices to allow such a thing. 

Once again sorry for refusing but hope you understand. 
Kind regards.

orazioedoardo:
@4s3ti Probably Mint gives a fatal error since it's not "Debian", "Ubuntu", "Raspbian" in the release file. Agree with the rest. At best there could be an argument like `--skip-unsupported` to force proceeding.

troygeiger:
@orazioedoardo There were no errors while installing; at least not for the OpenVPN option. I haven't tried the WireGuard option but that would probably work as well. Mint is based on Ubuntu and uses many of it package repos. All of the pivpn commands have been working without issue.
That being said, I understand the reasoning for not merging this pr. Supporting many distros would be a pain. I do like to idea of a --skip-unsupported though or maybe a --allow-unsupported.

Mihaib258:
Get the based of Linux Mint version you use and add it into the installation file if they are supported, for example for LM 19.3 as following:
	OSCN="bionic"
	PLAT="Ubuntu"
The file here contains this:
[install.sh.zip](https://github.com/pivpn/pivpn/files/4682986/install.sh.zip)
Please be aware not to use his file in other environment!
A better way is to modify the installer by adding "LinuxMint" and "tricia" version all over the file side by side with "Ubuntu" and "bionic".

Audioslaver:
As a user : making a VPN install script in order to simplify the VPN installing process is a neat progress...and definitely "fitted" for desktop distro users that dont want to spend hours.
My experience is that I have it installed on a dedicated machine at home with a Ubuntu LTS installed on it. I dont plan to install a Debian since I dont find it "Out of the box" enough. What I want is to install a distro and use it to do what I have to do with it, not spend hours installing it.
Mint has my preference over Ubuntu and has a significant share in terms of users now, and it would seem relevant for me to make piVPN run on it.
Power users dont need a script to run a VPN. People who are definitely happy to use PiVPN are noobs like I am, who are likely to use a distro like Mint.

PR #1048, opened 2020-05-14
Both wireguard and openvpn can be installed together (Issue #968)

shelleycat485:
Second protocol can be installed by doing another install, and selecting the protocol. Uninstall from dual detects if both are installed, and puts it back to using one protocol.  Tested on pi and AWS ubuntu.  Bash completion not used if two protocols are installed.

Readme changed to add that dual can be done.


orazioedoardo:
Cool, I tested the pull request and seems to work well, installation, management and uninstallation, with no conflicts. However there are some minor bugs:

1) There are still references to the single `/etc/pivpn/setupVars.conf` inside the self check script and the debug script.
2) Indentation [here](https://github.com/pivpn/pivpn/pull/1048/files#diff-409efba6db37b831b75997c11a4c9ec4R19-R30) and [here](https://github.com/pivpn/pivpn/pull/1048/files#diff-dcaa0a22b76e30da2cfe0d10e94603aeR22-R32) is too off.
3) When you install the pivpn script, you create a [symlink](https://github.com/pivpn/pivpn/pull/1048/files#diff-299d9b30c952310aace4aba77c698874R2246) of the VPN specific script to `/usr/local/bin/pivpn`, however you also do this for the second VPN, and ln will complain because the destination already exists. Not that big of a problem as you later delete the link and replace with the common pivpn script.
4) The uninstall script always prompts to choose which VPN you want to remove, even if you only have one.

...and things I (personally) would change:

1) It's not clear that you can install a second VPN using 'Reconfigure'.
2) I would use `pivpn ovpn` (as suggested in the FR) instead of `pivpn opv`.
3) Regarding bug 3, I would always use the common script even if you have a single VPN, and if the script detects two VPNs, it launches the specific pivpn.

Lastly, your pull request seems to use the master branch as reference. Currently master and test branches don't differ much, and there are no merge conflicts if I merge this. Though next time use the test branch as reference.

I wanted to make some updates to pivpn as well but I'll merge this one first to avoid conflicts. Tell me if you want to push more commits or settle on this.

shelleycat485:
Can you merge this one, and I will then work on the things you stated.  Thanks for reviewing it.

PR #1054, opened 2020-05-23
Tidy dual VPN protocol install

shelleycat485:
remove refs to /etc/pivpn/setupVars in selfcheck and debug scripts
  tidy indenting
  on install, check if symlink already exists before making one to avoid error
  uninstall indicates which vpns are available for uninstall
  selfcheck checks both protocols if both present
  install - additional text in reconfigure saying 2nd protocol can be added
  change to use pivpn ovpn insteed of pivpn opv when dual protocols exist
  change to README.md, since dual protocols not a major feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,6 @@
 About
 -----
 
-TS: 20/5/2020 starting change
-
-Modification that allows installing both openvpn and wireguard on the same machine.  To do this, run the installer twice locally, makeing sure you are going to pull branch test.  
-
-
 Visit the [PiVPN](https://pivpn.io) site for more information.
 This is a set of shell scripts initially developed by **@0-kaladin** that serve to easily turn your Raspberry Pi (TM)
 into a VPN server using two free, open-source protocols:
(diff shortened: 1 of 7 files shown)


orazioedoardo:
Why do you run the self check for both protocols at once?

shelleycat485:
Because if both are installed, why not - its possible that an error in one might have an effect on another; also part of the diagnostics should be to see what is currently installed.

orazioedoardo:
I find it to be confusing and not consistent to have both in the same place. I mean, if i run `pivpn ovpn -d` I want to see if openvpn is ok, not wireguard.

shelleycat485:
OK I can change that.  However, if you install both openvpn and wireguard listening on the same port; and if you set up different DNS providers for each of them they might interfere.  Should I just put out a statement that the other one is also installed - I would prefer to do that.

orazioedoardo:
> However, if you install both openvpn and wireguard listening on the same port;

Well, this should be checked at install time against any port a user might be using, not just when using two VPNs, but maybe this is a different pull request. 

> and if you set up different DNS providers for each of them they might interfere.

Mind explaining how? They can be different or be the same, what's the issue? It's just an outgoing connection to a server.

> Should I just put out a statement that the other one is also installed - I would prefer to do that.

The fact that `pivpn something` runs the help telling the user to use protocol specific syntax should be enough.

shelleycat485:
I will make a change to do only the single protocol from the pivpn xxx command.

PR #1056, opened 2020-05-25
Add extra ccd options to static host configuration file

twisniew1976:
For future will be good to add parameter to postpone adding extra ccd options.


orazioedoardo:
Let's wait for any interesting ccd option first.

PR #1085, opened 2020-07-05
Allow index based selection for WireGuard/QR code display

gizmocuz:
As we could have many client profiles, each with complex names, I thought it would be great if we could also enter a Index value instead of the profile name for the QR code display and delete option

For example:

:: Client list ::
• [1] HomeRob
• [2] HomeRobMobile
• ....
Please enter the Index/Name of the Client to show:

Where the user can either enter '2', or 'HomeRobMobile

I made a PR to test/illustrate this.

If you like the idea please let me know and I will try to implement this for delete and also for OpenVPN

Also feel free to implement this yourself if you like the idea as I am not a bash expert ;)

### DIFF
--- a/scripts/wireguard/qrcodeCONF.sh
+++ b/scripts/wireguard/qrcodeCONF.sh
@@ -33,17 +33,17 @@ if [ ! -s clients.txt ]; then
     exit 1
 fi
 
+LIST=($(awk '{print $1}' clients.txt))
 if [ "${#CLIENTS_TO_SHOW[@]}" -eq 0 ]; then
-
     echo -e "::\e[4m  Client list  \e[0m::"
-    LIST=($(awk '{print $1}' clients.txt))
+    len=${#LIST[@]}
     COUNTER=1
-    while [ $COUNTER -le ${#LIST[@]} ]; do
-        echo "• ${LIST[(($COUNTER-1))]}"
+    while [ $COUNTER -le ${len} ]; do
+        printf "• [%0${#len}s] %s\r\n" ${COUNTER} ${LIST[(($COUNTER-1))]}
         ((COUNTER++))
     done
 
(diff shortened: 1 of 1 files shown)


4s3ti:
@gizmocuz Thank you a lot for your contribution ... however, I have to say at the moment if this was up to me  .. up to me PR would be rejected without even reading its code or anything on it, please read:  https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md

Also It doesn't have any description on what its changing and why it would be good to add it. 

will leave the ultimate decision to @orazioedoardo since he has been taking care (and quite well!) of the repo lately!

gizmocuz:
Sorry about that, this was certainly not my intention. I will modify the description.

4s3ti:
@gizmocuz It's okay, however you are still missing the most important part :) 

You want to merge into master, we only accept pull requests to test branch. :)

gizmocuz:
@4s3ti , oeps you're right... If the 'idea' is OK then I will make a new PR where this is also done for 'delete' and for OpenVPN
The reason why this happens is that the 'default branch' here is the 'master' branch, and not a 'development/test' branch
I guess this will happen a lot
Happy to make a new PR, just let me know if you like the 'index' idea

orazioedoardo:
I think it makes sense since it’s faster than typing the name, however it should be consistent and work with add remove revoke, both openvpn and wireguard.

Also this would require a minimum length for the name like makeOVPN does otherwise you can’t tell if 1 is a name or the first client.

gizmocuz:
I have placed the Index values between brackets [ ] like in the first message.
Padding (left space(s)) is calculated based on the amount of clients

I will create a patch for remove and qr for both VPN and WireGuard. Hope to deliver this before the end of the weekend

Issue #1086, opened 2020-07-05
[FR] Index based selection for QR and delete


gizmocuz:
As we could have many client profiles, each with complex names, I thought it would be great if we could also enter a Index value instead of the profile name for the QR code display and delete option

For example:

::  Client list  ::
• [1] HomeRob
• [2] HomeRobMobile
• ....
Please enter the Index/Name of the Client to show:

Where the user can either enter '2', or 'HomeRobMobile

I made a PR to test/illustrate this.

If you like the idea please let me know and I will try to implement this for delete and also for OpenVPN

Also feel free to implement this yourself if you like the idea as I am not a bash expert ;)

https://github.com/pivpn/pivpn/pull/1085

PR #1087, opened 2020-07-08
Feature: Add Index based option for remove/qr commands

gizmocuz:
As we could have many client profiles, each with complex names, I thought it would be great if we could also enter a Index value instead of the profile name for the QR code display and delete option

For example:

:: Client list ::
• [1] HomeRob
• [2] HomeRobMobile
• ....
Please enter the Index/Name of the Client to show:

Where the user can either enter '2', or 'HomeRobMobile

This PR has been tested with OpenVPN/remove and WireGuard remove/qr commands

Previous version displayed a middle-dot at the beginning of each line, but i noticed a printing issue in one of my test systems and decided to remove it.
(It looked better with the dot)

https://github.com/pivpn/pivpn/issues/1086

### DIFF
--- a/scripts/openvpn/removeOVPN.sh
+++ b/scripts/openvpn/removeOVPN.sh
@@ -52,22 +52,32 @@ if [[ -z "${CERTS_TO_REVOKE}" ]]; then
         STATUS=$(echo "$line" | awk '{print $1}')
         if [[ "${STATUS}" = "V" ]]; then
             NAME=$(echo "$line" | sed -e 's:.*/CN=::')
-            CERTS[$i]=${NAME}
             if [ "$i" != 0 ]; then
                 # Prevent printing "server" certificate
-                printf "  %s\n" "$NAME"
+                CERTS[$i]=${NAME}
             fi
             let i=i+1
         fi
     done <${INDEX}
+    
+    i=1
+    len=${#CERTS[@]}
+    while [ $i -le ${len} ]; do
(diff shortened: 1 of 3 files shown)


orazioedoardo:
I think `1) HomeRob` resembles a list better then `[1] HomeRob`.

gizmocuz:
Made the changes to display it in a better format

orazioedoardo:
> Made the changes to display it in a better format

I see, thanks! Will test later.

orazioedoardo:
As I said [here](https://github.com/pivpn/pivpn/pull/1085#issuecomment-655492023), I think there should be a minimum lenght requirement for names, otherwise:

```
$ pivpn -r
::  Client list  ::
1) test
2) 3
Please enter the Index/Name of the Client to be removed from the list above: 3
:::  does not exist
```

You can't remove client '3' by name because '3' is an index and the code will delete the third client (if it exists).

gizmocuz:
@orazioedoardo , I agree, but that is not related to this patch. We could make another patch for the name requirements.
I also think you should enter a good name, not 1,2,3, we're asking to enter a name, not a number.
It is also very bad practice to name things 1,2,3 instead of RemoteMobile, RemoteLaptop, Remote....

orazioedoardo:
Sure, '3' is not a good name, however I think for correctness the code should handle those cases, for example require minimum length of 4 so 1111 is definitely a name (you can't have more than 253 clients).

I think it's related to this patch because technically you could remove a client named '3' and now you can't in certain cases. However I'm going to make some changes to the test branch so maybe I can do that myself.

PR #1091, opened 2020-07-11
add parameter to force remove profile

gi8lino:
This PR adds a -f|--force parameter to delete a profile without a prompt (useful if you deploy it with ansible).

### DIFF
--- a/scripts/wireguard/removeCONF.sh
+++ b/scripts/wireguard/removeCONF.sh
@@ -30,7 +30,7 @@ do
             helpFunc
             exit 0
             ;;
-        -yes|--yes)
+        -y|--yes)
             CONFIRM=true
             ;;
         *)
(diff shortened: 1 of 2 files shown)


orazioedoardo:
Two requests:

1) Change `-f|--force` with `-y|--yes` as force implies something that shouldn't be done but the user wants to do anyways. -y aligns with most utilities too.
2) Add the same confirmation before revoking, as well as -y to the removeOVPN.sh script.

gi8lino:
I have implemented both requests

scripts/wireguard/removeCONF.sh
```diff
@@ -29,6 +30,9 @@ do
             helpFunc
             exit 0
             ;;
+        -yes|--yes)
```
orazioedoardo, line 33:
```suggestion
        -y|--yes)
```

Typo

PR #1101, opened 2020-07-31
Delete/cleanup private key and profile, use shred if available

fa18swiss:
As there file is not used after and they contains the private key, I think they can be deleted.

### DIFF
--- a/scripts/openvpn/makeOVPN.sh
+++ b/scripts/openvpn/makeOVPN.sh
@@ -34,6 +34,20 @@ helpFunc() {
     echo ":::  -h,--help            Show this help dialog"
 }
 
+safeDelete() {
+    if [ -z "$1" ]; then
+        exit
+    fi
+    if [ ! -f "$1" ]; then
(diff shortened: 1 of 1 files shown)


orazioedoardo:
The private key is actually used by `pivpn -o` to regenerate a configuration if the user needs to. Even if it wasn't used, what's the benefit in removing the file?

[shred](https://man7.org/linux/man-pages/man1/shred.1.html) is not useful because on SD Cards (and flash memory in general) there is no guarantee that when writing over a file you overwrite the same physical block. Actually it's likely that the physical block changes on every write due to wear leveling, to distribute writes all over the memory instead of a single flash chip.

Also, pull requests should go to the test branch: https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md#do-you-want-to-open-a-pull-request-to-modify-the-source

fa18swiss:
I have created a new PR ( #1104 ) taking into consideration your remarks.

PR #1104, opened 2020-08-03
Delete profile and add option to delete private key

fa18swiss:
Remove the local copy of the profile.
Add option -r | --remove to delete private key (after that the option -ovpn will not be possible)

Why deleting these files :
- The profile can be regenerated anytime with -o option, so the file is not useful
- Delete private key is a good security practice
-- If you lost an .ovpn profile, it is a better security practice to revoke it and generate a new than use it again
- Low disk usage (but those few kBs aren't relevant)


orazioedoardo:
Sorry but what benefit you get from removing the private key? If someone gets the key it can impersonate a client, however to get the key you need root access and at that point the system is already compromised. Also, the same private key is inside the .ovpn file in the home directory of the user.

orazioedoardo:
Closing due to inactivity.

PR #1149, opened 2020-09-22
Handles special characters better

xptsp:
These changes allow PiVPN scripts to display and revoke certificates with common names like "José" better.  Prior to this commit, names like "José" would be shown as "Jos\xC3\xA9".

### DIFF
--- a/scripts/openvpn/listOVPN.sh
+++ b/scripts/openvpn/listOVPN.sh
@@ -20,11 +20,11 @@ while read -r line || [ -n "$line" ]; do
     EXPD=$(echo "$line" | awk '{if (length($2) == 15) print $2; else print "20"$2}' | cut -b 1-8 | date +"%b %d %Y" -f -)
         
     if [ "${STATUS}" == "V" ]; then
-        printf "Valid  \t  %s  \t  %s\\n" "$NAME" "$EXPD"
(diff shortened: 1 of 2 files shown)


orazioedoardo:
> Prior to this commit, names like "José" would be shown as "Jos\xC3\xA9".

Maybe it's due to your terminal encoding? Works for me.

```
$ NAME="José"
$ printf ""$NAME"\n"
$ José
```

xptsp:
I tried that in a BASH shell.  It worked just fine.  But the PiVPN scripts seem to manage that name (and others, I assume).  This commit fixes the issue for displaying it, as well as revoking that certificate.

orazioedoardo:
Sorry for the delay, you were right, however yo need to change the target branch to test, according to https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md

xptsp:
How do I do that?

orazioedoardo:
https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request

You also need to add additional commits since listOVPN.sh in the test branch was different when you originally submitted the pull request.

xptsp:
Okay, I'm closing this pull request and will open another one in the right branch.  Thank you!

PR #1164, opened 2020-10-20
use wg syncconf

jeffrysurya:
thank you for your amazing pivpn project. i just read documentationon wg man page, and i found how to reload wg configuration without interrupt existing connection. by change systemctl restart wg-quick@wg0 to wg syncconfig wg0 <(wg-quick strip wg0). i already implement this on my server and works like charm, and i add another option to manual restart wireguard. thank you

Issue #1111, opened 2020-08-09
A feature to disable / enable single wireguard client configs would be very apreciated 


delight:
It would be very nice to be able to disable / enable a wireguard client configuration

In that way you can disable / enable an affiliated account as needed in a certain situation - resulting in more flexibility.

Possible applications:
- Give a client access on request
- Enable disable access on certain times only ( with a little crontab )
- ... I bet there is even more applications to this

shelleycat485:
Starting to look at this.  My plan is to:
Code like the delete, but comment out lines for the user in configs, wg0.conf instead of delete.
Add a (new) log file to the configs directory giving a timestamp of when the disable or re-enable takes place.
Have the option of having the user on the command line, so that a cron job can easily enable/disable
Check the add user so that a commented out user is considered for clashes when adding a new user
Check the add user so that a commented out user address (10.6.0...) is not reused
Check the delete user so that a commented out user can also be deleted
Check the -c command so that disabled users are listed as well
Leave the .conf file for the client in place whether disabled or not.
If someone else is looking at this, please let me know.

shelleycat485:
Still looking at this.  At the moment cannot test on AWS 20 or AWS 18 because cannot install - fails to find iptables-persistent package!.   I've tried installing from the test branch of pivpn directly, with no other changes, and I get the same.

4s3ti:
Merged into master: https://github.com/pivpn/pivpn/commit/b5ec995f68fbae2eb30b50e1b6a4350f96316f2e

delight:
Is there going to be a new release and will it be necessary to reinstall?

Thnx

4s3ti:
@delight pivpn does not work with releases ... this is an installer and it has done its job after you have used it ... you can update PiVPN Scripts by running pivpn again and it will offer you to update them.

delight:
Thank you for clarifying 👍🏻

PR #1166, opened 2020-10-22
Adding temporary Disable and Enable wireguard clients

shelleycat485:
Adds pivpn disable or -dis, and pivpn enable -ena.  This disables or re-enables an existing wireguard client.   pivpn -dis -v will show which clients are disabled.  pivpn -ena -v also show which clients are disabled.   Also pivpn -l will indicate if any clients are disabled.  This is implemented by changing the wg0.conf file only, no other configuration files are changed.

Testing of disable followed by disable and enable followed by enable and on non-existent clients seem OK.
Disabled clients can be removed. A pivpn add does not work if a disabled client with the same name exists. 
At this moment this does not work on AWS ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-* because the wireguard service cannot be reloaded (see pull request #1164 ).  It has been otherwise tested on that machine, and on a RPi4, Buster.  Addresses issue #1111.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -13,8 +13,8 @@
 # timestamp 2020/5/24 15:53BST
 
 ######## VARIABLES #########
-pivpnGitUrl="https://github.com/pivpn/pivpn.git"
-#pivpnGitUrl="/home/pi/repos/pivpn"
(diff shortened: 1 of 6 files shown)


shelleycat485:
withdrawing, will reapply and retest with d860f1d applied

PR #1174, opened 2020-11-10
Update makeCONF.sh

### LINKED ISSUE #990: PersistentKeepalive Client
**Is your feature request related to a problem? Please describe.**
A PersistentKeepalive should be added automatically when creating a client, I use PersistentKeepalive 1

**Describe the solution you'd like**
Add PersistentKeepalive to the script

franklouwers:
Add PersistentKeepalive parameter to client config, closes #990

### DIFF
--- a/scripts/wireguard/makeCONF.sh
+++ b/scripts/wireguard/makeCONF.sh
@@ -110,6 +110,7 @@ echo "[Peer]
 PublicKey = $(cat keys/server_pub)
 PresharedKey = $(cat "keys/${CLIENT_NAME}_psk")
 Endpoint = ${pivpnHOST}:${pivpnPORT}
+PersistentKeepalive = 25
 AllowedIPs = 0.0.0.0/0, ::0/0" >> "configs/${CLIENT_NAME}.conf"
 echo "::: Client config generated"
 


orazioedoardo:
I don't this the persistent keepalive should be enabled by default. I think the target audience of PiVPN is mostly made up of users new to VPNs, which are more likely to use it to connect a device to their network than the other way around (never seen an open issue about that). Given that, persistent keepalive would prevent the device from going to sleep, affecting battery live.

TinCanTech:
This is technically no different to using --keepalive in Openvpn, which you do use.

orazioedoardo:
> This is technically no different to using --keepalive in Openvpn, which you do use.

Forgot about that. Does removing `--keepalive` mean an increased overhead when sending packets due to a full TLS handshake going on?

TinCanTech:
> Does removing `--keepalive` mean an increased overhead when sending packets due to a full TLS handshake going on?

Openvpn `--keepalive` is a macro to make configuring `--ping`/`--ping-restart`configurable for all clients from the server side in a simple way.

However, your question leads to two answers:
* If you don't have any `--keepalive` or `--ping*`settings then once connected, the VPN will simply sit and wait for data. It will most probably do a renegotiate at one hour (`--reneg-sec` default).

* If you don't have `--keepalive` but do have appropriate `--ping*` settings then the VPN should function normally.

`--keepalive` is recommended for all but the _most-corner_ cases, even devices with limited battery life.

This is one of those questions that can be argued either way but from Openvpn _point of view_ `--keepalive` has solved many more problems than it has ever caused.

orazioedoardo:
> If you don't have any `--keepalive` or `--ping*`settings then once connected, the VPN will simply sit and wait for data. It will most probably do a renegotiate at one hour (`--reneg-sec` default).

In such case, if both client and server are idle, NAT expires and then a clients wants to send data again, would it renegotiate again? Or "reuse" the existing session straight?

TinCanTech:
For Openvpn, in most cases the client would `float` to a new source address, generally only new source port number in the case of NAT expiry.

orazioedoardo:
> For Openvpn, in most cases the client would `float` (No TLS negotiation) to a new source address, generally only new source port number in the case of NAT expiry .. That is, provided the packet passes all authentication tests.

Given that, it may sound that we don’t really need keepalive or ping*, or there are other reasons why in general is best to use it as you suggested above?

TinCanTech:
I can see why you like the idea but given how long PiVPN has used --keepalive and the number of users changing it would effect, I would not be comfortable making that change without reasonable testing first.

Perhaps this could be a configurable option for PiVPN:
* Roadwarrior only - No `--keepalive` required.
* Remote client-LAN access - Use --keepalive` (Recommended)

In theory it should work fine.

TinCanTech:
I have done some testing with current Openvpn and I would **not** recommend removing `--keepalive`.  Using UDP, Openvpn relies quite heavily on timers/timeouts and without `--keepalive` TLS keys (eg. `--tls-crypt`) fail to initialise.  I know that Openvpn does have some issues to do with connecting using TLS but they have not been resolved yet (https://community.openvpn.net/openvpn/ticket/1310#comment:2)

orazioedoardo:
Thanks for the explanations, better keep it then.

franklouwers:
As you would keep it for OpenVPN, can we discuss adding it for WireGuard?

orazioedoardo:
@franklouwers I'd say no for the reason above: https://github.com/pivpn/pivpn/pull/1174#issuecomment-725421987.
It's a minor benefit for unusual use case (server -> client) with a default drawback (constant "ping").

orazioedoardo:
Closing as inactive.

Issue #1204, opened 2020-12-14
[BUG] UPDATE_PKG_CACHE seems to not work


4s3ti:
## In raising this issue, I confirm the following: 

`{please fill the checkboxes, e.g: [X]}`

- [x] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*.
- [x] The issue I am reporting can be *is* directly related to the pivpn installer script.
- [] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pivpn/pivpn/wiki/FAQ), [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen)).

@orazioedoardo First of all i am sorry i am raising duplicated stuff that has been talked before, but as you know been in Lurk mode, but seen this happening quite often, and with all the testing i am trying maybe it will be possible to find some clues on how to fix this. 

**Describe the bug**

Has we have seen in the past, some users complain that the script fails then they run apt udpate and it works, 
During testing i stumbled upon this scenario, on ubuntu 18.04, 
2 different systems, same distro, one is a VPS that i spun up for some manual testing, the another one is the VM that travis spins up to run the testing. 
on the VPS the script worked straight away, 
on Travis I need to add `apt update` as first step of the script, otherwise `iptables-persistent` would fail to install 
This doesn't seem to have happened in any scenario on ubuntu 16.04 so far. 

```
:::    Checking for whiptail... already installed!
:::    Checking for net-tools... already installed!
:::    Checking for bsdmainutils... already installed!
:::    Checking for iptables-persistent... not installed!
:::    Failed to install iptables-persistent!
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package iptables-persistent
```

**To Reproduce**
Not sure how to reproduce and on which conditions this happens.

**Expected behavior**
PiVPN should update package cache which seems to not happen in some situations

**Additional context**

Some questions that it brings me... 

Is `updatePackageCache()` really working all the time? what could cause it not to work some times?

**Possible solution 1**

[...]

PR #1223, opened 2021-01-12
Fix for #1204

4s3ti:
Added -y to $UPDATE_PKG_CACHE
updatePackageCache() no longer checks if apt update was run, it will
always update package cache since its a requirement
Replaced all updates using ${UPDATE_PKG_CACHE} with
updatePackageCache()


4s3ti:
@orazioedoardo when you can please take a look if you agree with the fix ... 

Temporary tests i am working at can be seen here: https://travis-ci.com/github/4s3ti/pivpn/builds/212849852

Issue #1231, opened 2021-01-23
[BUG] pivpn script does not return proper exit code


glitch452:
## In raising this issue, I confirm the following: 

- [x] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*.
- [x] The issue I am reporting *is* directly related to the pivpn installer script.
- [x] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pivpn/pivpn/wiki/FAQ), [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen)).

**Describe the bug**
When there is an error which should cause the pivpn script to produce a non-zero exit code, 0 is returned instead.  

When running the following command: `pivpn revoke --yes michael123`
The following output is produced: `You passed an invalid cert name: 'michael123'!`
Then, when running `echo $?` the output is `0`

**To Reproduce**
1. Run `pivpn revoke --yes michael123`
2. Then Run `echo $?`
3. The contents of `$?` should be `1` not `0`

**Expected behavior**
The exit code from the script in this context should be `1`

**Desktop (please complete the following information):**
 - VPN Protocol: Open VPN
 - OS: Raspbian buster

**Additional context**
I searched for the error message and found it in `pivpn/scripts/openvpn/removeOVPN.sh`.  Lines 120-121.  There is clearly an `exit 1` statement, but the pivpn script is not passing along the exit code. 

In the file `pivpn/scripts/openvpn/pivpn.sh` the function that loads `removeOVPN.sh` is on line 45.  On line 46 the exit code is hard coded to `exit 0` instead of exiting with the exit code of the called script.  Perhaps all those exits in the main script that follow a sub-script should be `exit $?` to pass along the exit codes.

glitch452:
See PR #1232

4s3ti:
Merged to master 7fdbe44df5319b7c99a4481c754acd8c0c6c98fa closing.

PR #1232, opened 2021-01-23
Fix exit codes when running a secondary script

glitch452:
When running a secondary script, the exit code of the main script will use the exit code from the secondary script instead of always returning 0.  That way errors can be detected by the exit code when using the main pivpn script.  

Also, fixed an exit code when running install.sh without sudo to return 1 instead of 0.  

This is related to Issue #1231

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -143,7 +143,7 @@ main(){
 		:
 	elif [ "$UpdateCmd" = "Update" ]; then
 		$SUDO ${pivpnScriptDir}/update.sh "$@"
-		exit 0
+		exit "$?"
 	elif [ "$UpdateCmd" = "Repair" ]; then
 		# shellcheck disable=SC1090
 		source "$setupVars"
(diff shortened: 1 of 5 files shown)


TinCanTech:
Note: `function debugFunc` occurrences  will still always `exit 0`

glitch452:
> Note: `function debugFunc` occurrences will still always `exit 0`

Looks like it can return exit 1 if the setup vars file is missing: `pivpnDebug.sh` lines `6-9`
```bash
if [ ! -f "${setupVars}" ]; then
    echo "::: Missing setup vars file!"
    exit 1
fi
```

TinCanTech:
It is not a problem .. just a note.

4s3ti:
Hi @glitch452 Thanks a lot for your contribution,  I have actually noticed this quite a while ago just didn't had the opportunity to take care of it yet ... It looks good to me .. merging.

Issue #1233, opened 2021-01-26
[Improvement] Make ShellCheck Happy


4s3ti:
PiVPN is currently failing on Linting tests, 

https://travis-ci.com/github/pivpn/pivpn/jobs/475625045

PiVPN scripts should pass on shellcheck and this should be later used to evaluate PR quality. 

Since it is not Breaking anything in the script, I have added `allow_failures` for linting stage until this has been closed.

orazioedoardo:
We should also have a coding style guide. As of now coding style is not consistent.

4s3ti:
agree! but i am not sure if am the right one to implement that .... having shellcheck checking for good practices is already a good start.

orazioedoardo:
We could just link to this one: https://google.github.io/styleguide/shellguide.html

DundarGoc:
I've had the most success with [shfmt](https://github.com/mvdan/sh) when it comes to shell formatters and can highly recommend it.

DundarGoc:
I'm new to pull requests, do you prefer one mid-sized pull request with a list of of changes och one pull request per change?

4s3ti:
@DundarGoc Personally i prefer smaller fixing one issue at a time as they tend to be easier to read and review, also makes easier to track any issue in case it comes up :) this said i don't mind having a thouzand of commits for the same issue :)

DundarGoc:
I can't find setupVars.conf in the repo, is it created at runtime or something?

4s3ti:
@DundarGoc Yes, basically during runtime script echoes the variables and its values to that file which is in turn used later by the pivpn scripts as a way to identify which options were used during setup and by the --debug option to info on whichsettings were used during the setup

That file can also be used for `--unattended` setups ... 1 example here: https://github.com/pivpn/pivpn/blob/master/ciscripts/ci_wireguard.conf

4s3ti:
Chellsheck is now happy on test branch! 

CI Job will no longer tolerate failures on Lint/shellcheck stage.

4s3ti:
Merged to master today. 
https://github.com/pivpn/pivpn/commit/d4e3a63522e4edaa2f9ad9f2d8dd6ffab4117ff9

Closing ...

PR #1241, opened 2021-02-07
Fixed shellcheck warnings SC2219, SC2006, SC2004 and SC2129. Issue #1233

DundarGoc:
SC2219: "Instead of 'let expr', prefer (( expr ))."
SC2006: "Use $(...) notation instead of legacy backticked `...`."
SC2004: "$/${} is unnecessary on arithmetic variables."
SC2129: "Consider using { cmd1; cmd2; } >> file instead of individual redirects."


4s3ti:
Generally Looks good to me thanks a lot! :D however please get the merge request to show only your ... don't really understand why it is throwing also all those ones i have done.

DundarGoc:
No clue why the other commits came up. One suspicion is that I branched from master and not test. I'm gonna try again and see if it changes.

PR #1242, opened 2021-02-07
Fixed shellcheck warning SC2004, SC2006, SC2129, SC2219. Issue #1233.

DundarGoc:
SC2004: "$/${} is unnecessary on arithmetic variables."
SC2006: "Use $(...) notation instead of legacy backticked `...`."
SC2129: "Consider using { cmd1; cmd2; } >> file instead of individual redirects."
SC2219: "Instead of 'let expr', prefer (( expr ))."


DundarGoc:
Do you want one error code per pull request or multiple error codes in one go?

Issue #1239, opened 2021-02-03
Need to install wirguard before running PiVPN


sidewinder94:
## In raising this issue, I confirm the following: 

`{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting *is* directly related to the pivpn installer script.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pivpn/pivpn/wiki/FAQ), [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen)).

**Describe the bug**

On Debian 10 (ARM64) the install script does not ask me to choose which VPN service to install

**To Reproduce**

1. Clean Debian 10 (ARM64) installation
2. Run PiVPN installer script
3. Follow the steps

**Expected behavior**
The installer asks me to choose which VPN system I want to use and installs it.

**Desktop (please complete the following information):**
Not relevant, server installation does not complete

**Smartphone (please complete the following information):**
Not relevant, server installation does not complete

**Actual behaviour**
I do not get asked to choose the VPN service I want to use and the setups directly goes to the unattended upgrade setup.

**Additional context**

After the installer finished the following lines can be seen 

```
::: Checking for existing base files...
:::    Checking /usr/local/src/pivpn is a repo... OK!
:::     Updating repo in /usr/local/src/pivpn... done!
main: line 1140: [: : integer expression expected
main: line 1152: [: : integer expression expected
::: Install Complete...
```

4s3ti:
Dear user, even though we love the fact you are using pivpn, the maintainers of this project keep answering the same thing over and over again.

They keep spending time not only on the project that makes your life so great but also on documentation that you not to read at all therefore this issue is being closed. 

Don't know why your issue got closed? 

You can start by reading here: https://www.pivpn.io/#contrib
And here: https://github.com/pivpn/pivpn/blob/master/README.md
And here: https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md

and the first lines on the issue template that appeared after you have pressed all those green buttons to open up this issue: 

```
<!-- 
# PiVPN Issue Template
[...]

PR #1245, opened 2021-02-11
Default VPN protocols to unsupported (workaround for issue #1239)

orazioedoardo:
From my comment: https://github.com/pivpn/pivpn/issues/1239#issuecomment-777427001

> By the way since Bullseye is going to ship with kernel 5.10 with built-in WireGuard, I think it makes sense to just fix the script for now to stop the installation on arm64 Debian 10.

OpenVPN should be offered to install then.


4s3ti:
lgtm

PR #1249, opened 2021-02-13
Fixed shellcheck issue SC2015

DundarGoc:
SC2015: Note that A && B || C is not if-then-else. C may run when A is true.

Replaced `[ $1 -gt 1 ] && shift $1 || shift` with `shift $1` since
`shift 1` is identical to `shift`.


TinCanTech (approved):
lgtm

4s3ti:
Merged, Thank you :D

Issue #1273, opened 2021-03-08
[FR] Allow to specify `pivpnNET`, `subnetClass` and `ALLOWED_IPS`


DerDanilo:
**Is your feature request related to a problem? Please describe.**
Kind of. It is not possible to specify custom `pivpnNET`, `subnetClass` and `ALLOWED_IPS` without manual interaction in multiple ways.

**Describe the solution you'd like**
It should be possible to specify a custom vpn `network` and `subnet` as well as `allows ips` if the vpn is used to only route specifiy subnets/hosts through it.

- Read `pivpnNET`, `subnetClass` and `ALLOWED_IPS` from unattend setup file (currently ignored)
- Ask if custom `pivpnNET`, `subnetClass` and `ALLOWED_IPS` are desired in setup prompts

**Describe alternatives you've considered**
Manually setting up WG or using other solutions that allow customizing them upfront.

**Additional context**
Please ask if more info is required.

orazioedoardo:
> Read pivpnNET, subnetClass and ALLOWED_IPS from unattend setup file (currently ignored)

Can be done.

> Ask if custom pivpnNET, subnetClass and ALLOWED_IPS are desired in setup prompts

Asking for pivpnNET and subnetClass with prompts in a beginner script is overkill IMHO, ALLOWED_IPS yes but not as is but something like that:

- Forward all traffic through the VPN
- Forward only traffic directed to the local network

DerDanilo:
pivpnNET needs to be different sometimes. Hence there could be a similiar prompt like:

- Use default VPN network (10.6.0.0/24)
- Configure custom VPN network

I think this is an important config to be available to configure.

It would be of help if this would be configurable via the unattend setup file.

Thanks

DerDanilo:
Here we go regarding the setupVARs part. This is only concerning the setupVARs file config. GUI prompts still have to be added if required (not by me).

https://github.com/pivpn/pivpn/pull/1279

4s3ti:
Merge request related with this has been merged into master. Closing.

PR #1279, opened 2021-03-15
Support setupVars for pivpnNET, subnetClass and ALLOWED_IPS via unattended setup

DerDanilo:
Support setupVars for `pivpnNET`, `subnetClass` and `ALLOWED_IPS` via unattended setup.

- **OpenVPN** --> `pivpnNET` and `subnetClass`
- **Wireguard** --> `pivpnNET`, `subnetClass` and `ALLOWED_IPS`

Fix part of https://github.com/pivpn/pivpn/issues/1273

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1093,9 +1093,12 @@ installPiVPN(){
 			pivpnNET="10.6.0.0"
 		fi
 		vpnGw="${pivpnNET/.0.0/.0.1}"
-		# Forward all traffic through PiVPN (i.e. full-tunnel), may be modified by
-		# the user after the installation.
-		ALLOWED_IPS="0.0.0.0/0, ::0/0"
+		# Allow custom allowed IPs via unattend setupVARs file. Use default if not provided.
+		if [ -z "$ALLOWED_IPS" ]; then
+			# Forward all traffic through PiVPN (i.e. full-tunnel), may be modified by
+			# the user after the installation.
+			ALLOWED_IPS="0.0.0.0/0, ::0/0"
+		fi
 		CUSTOMIZE=0
 
 		installWireGuard
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1063,7 +1063,10 @@ installPiVPN(){
 	if [ "$VPN" = "openvpn" ]; then
 
 		pivpnDEV="tun0"
-		pivpnNET="10.8.0.0"
+		# Allow custom NET via unattend setupVARs file. Use default if not provided.
+		if [ -z "$pivpnNET" ]; then
+			pivpnNET="10.8.0.0"
+		fi
 		vpnGw="${pivpnNET/.0.0/.0.1}"
 
 		askAboutCustomizing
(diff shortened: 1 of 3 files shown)


DerDanilo:
I am not completely sure that it's early enough to define `subnetClass` in the function `installPiVPN()` since I removed it from the top of the script to have it at the actual place were it can be defined. This makes the check easier if it was pre-defined without the need to add another var.

```
	# Allow custom subnetClass via unattend setupVARs file. Use default if not provided.
	if [ -z "$subnetClass" ]; then
		subnetClass="24"
	fi
```

orazioedoardo:
Can't text the pull request(s) however I can tell that those lines are used to set the static IP settings when using unattended on Raspbian: https://github.com/pivpn/pivpn/pull/1279/files#diff-bb14cdae0728d24ba60e6b112a95f3fd4483e19424969f77308b4954348d5fc8L2-L4

DerDanilo:
> Can't text the pull request(s) however I can tell that [these](https://github.com/pivpn/pivpn/pull/1279/files#diff-bb14cdae0728d24ba60e6b112a95f3fd4483e19424969f77308b4954348d5fc8L2-L4) lines are used to set the static IP settings when using unattended on Raspbian.

What does "Can't text the..." mean?

Are those vars required when setting up wireguard VPN or not? I went through the code and checked all references. They only seem to be required in connection with OpenVPN. So not needed with Wireguard.

orazioedoardo:
> What does "Can't text the..." mean?

Test, actually.

> Are those vars required when setting up wireguard VPN or not? I went through the code and checked all references. They only seem to be required in connection with OpenVPN. So not needed with Wireguard.

Look at [this](https://github.com/pivpn/pivpn/blob/d7771c251418fa443869397d46f93c5b0c197558/auto_install/install.sh#L705) function.

DerDanilo:
Thanks. I checked this function before and could not see why those values should be required by wireguard. 

The genegrated setupVars.conf file does not contain them either. If I understand the `reconfigure=false` flag correctly the script should fail if those vars are required.

I can remove the commit regarding those vars. But I'd like to keep the example files correct if possible.

If I understand those vars correctly they allow to specify the local subnet, including mask and the default gw that the VPN server should use itself? DHCP is not required for Wireguard at all, that is why I am asking.

```
root@v:~# cat /etc/pivpn/wireguard/setupVars.conf
PLAT=Debian
OSCN=buster
USING_UFW=0
IPv4dev=eth0
install_user=wgadmin
install_home=/home/wgadmin
VPN=wireguard
pivpnPORT=1194
pivpnMTU=1280
pivpnDNS1=192.168.25.248
pivpnDNS2=192.168.25.249
pivpnHOST=v.mydomain.com
INPUT_CHAIN_EDITED=0
FORWARD_CHAIN_EDITED=1
pivpnPROTO=udp
pivpnDEV=wg0
pivpnNET=10.10.0.0
subnetClass=24
ALLOWED_IPS="10.10.0.0/24, 192.168.25.0/24"
UNATTUPG=1
INSTALLED_PACKAGES=(iptables-persistent qrencode qrencode iptables-persistent qrencode)
```

orazioedoardo:
> The genegrated setupVars.conf file does not contain them either.

Because you tested the script on Debian, it does on Raspian.

> If I understand those vars correctly they allow to specify the local subnet, including mask and the default gw that the VPN server should use itself?

Yes

Switch target branch to test please.

DerDanilo:
> Because you tested the script on Debian, it does on Raspian.

So if one should use those vars we can leave them in the example file. Maybe we should extend the documentation and explain what those vars are for. 



> Switch target branch to test please.

Done

DerDanilo:
How do we get this merged now? In my test installations it was working fine.

4s3ti:
Please resolve the conflicts on your base branch please, seems to be related with the merge of one of your merge requests.

DerDanilo:
> Please resolve the conflicts on your base branch please, seems to be related with the merge of one of your merge requests.

Apparently it was caused by the other merge request I pushed. Will check this soon and for it fixed.

DerDanilo:
Sorry for the delay. Was busy and then forgot about it. Hope it's fine now and can get merged soon.

PR #1303, opened 2021-05-06
Fix #1302

### LINKED ISSUE #1302: [BUG] pivpn.sh was named incorrectly
## In raising this issue, I confirm the following: 

- [X] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting *is* directly related to the pivpn installer script.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pivpn/pivpn/wiki/FAQ), [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen)).

**Describe the bug**
After installing, I could not use `pivpn`
image
I chased the symlink in `/usr/local/bin`
image
I realized the symlink was invalid and was pointing to /usr/local/src/pivpn/scripts/pivpn.sh
However, the `pivpn.sh` script was named `pivpn`
image
So, I renamed `pivpn` to `pivpn.sh`
image
-Symlinked is now fixed
image
-can now use `pivpn`
image

TLDR: 
Installer script had an issue: `/usr/local/src/pivpn/scripts/pivpn.sh` was named `pivpn` and not `pivpn.sh`
and this caused the pivpn symlink to not work

**To Reproduce**
Installed with `curl -L https://install.pivpn.io | bash`

**Expected behavior**
pivpn should work
`/usr/local/src/pivpn/scripts/pivpn.sh` should be named `pivpn.sh` and not `pivpn`

p1r473:
pivpn.sh and bash-completion.sh were missing the file extensions
https://github.com/pivpn/pivpn/issues/1302


4s3ti:
Hi, Thanks a lot for your PR however,

This PR Will break a lot in pivpn as you are only changing the extension and to change those you need to change all the references to those files in multiple places, its not just like "change the file extension" 

on Linux systems the extensions don't really matter for anything and there even though the way we install those scripts nowadays doesn't really matter much, in the past those scripts were like that with a purpose. 

All things considering I am not accepting this PR as it brings no benefit VS high risk of missing something and breaking it. 


If you believe there is a really good reason for doing so, please feel free to re-open and let me know which and ill consider it. 

Kind regards.

PR #1305, opened 2021-05-06
Fix corrupted QR code image in various fonts

GHPS:
(affected fonts include Ubuntu Mono, Mitra Mono, InconsolataGo Nerd Mono)

### DIFF
--- a/scripts/wireguard/qrcodeCONF.sh
+++ b/scripts/wireguard/qrcodeCONF.sh
@@ -60,7 +60,7 @@ for CLIENT_NAME in "${CLIENTS_TO_SHOW[@]}"; do
     if grep -qw "${CLIENT_NAME}" clients.txt; then
         echo -e "::: Showing client \e[1m${CLIENT_NAME}\e[0m below"
         echo "====================================================================="
-        qrencode -t ansiutf8 < "${CLIENT_NAME}.conf"
+        qrencode -t ansi256 < "${CLIENT_NAME}.conf"
         echo "====================================================================="
     else
         echo -e "::: \e[1m${CLIENT_NAME}\e[0m does not exist"


GHPS:
> Wondering if the ones working will break with this ... have you tested @GHPS ?

The ANSI-encoding should be supported by any font since it relies on ASCII-only characters.
During testing I encountered no problems with "-t ansi256"

4s3ti:
LGTM then .. merging. 

This should go into master branch tomorrow most likely as i might merge test into master.

GHPS:
Some background on this problem might be useful: On my machine (Kubuntu 18.04,
zsh with InconsolataGo Nerd Mono) the qr code generation (pivpn -qr) some day gave 
a distorted, unscanable picture while it previously worked fine without a problem.

The cause (fresh installation of said font) was easy to find, the original cause
not so. After some research I found that qrencode uses the Unicode Block Elements [1]
in the "utf8" options (UTF8, ANSIUTF8, ANSI256UTF8). These characters get combined
with the the ordinary Space character to form the image (which to the standard is
legit).

While this is the modern path to generate block graphics, various potholes can get
in the way:

1) The font don't have Unicode Block Elements.[2]
2) The Block Elements have some defects, e.g. are centered [3]
3) The Block Elements are wider than normal characters especially Space.
   Currently this is the case with Ubuntu Mono, Mitra Mono, InconsolataGo Nerd Mono.

When using qrencode as a CLI tool it is easy to test whether the font currently in
use supports the "utf8" option correctly and switch back to simple "ansi" in case
of a distorted image.

But for imbedded use of qrencdoe (like pivpn -qr) and users unfamiliar with
the aspects described above, the problem can become unsolvable. Therefore
it makes sense to use only "ansi" encoding even if the resulting qrcode is
larger and less pretty. BTW the margins can be reduced to 2 (option -m 2)
without decreasing the scan quality.


[1]: https://en.wikipedia.org/wiki/Block_Elements
[2]: https://github.com/fukuchi/libqrencode/issues/148
[3]: https://github.com/microsoft/terminal/issues/7264

GHPS:
> `qrencode -m 1 -M -t ansi256`

I had problems scanning code with -m 1 margin but -m 2 is fine.

4s3ti:
Fix on test branch :)

4s3ti:
Fix pushed to master branch.

GHPS:
> This change generates a terribly large QR-Code.

Yes - after thinking about this problem for a while I'm wondering if an additional option for 
pivpn would make sense. An option like -t after -qr, e.g.

`pivpn -qr -t ansiuft8`

This way a user can decide whether to use the extended character encoding or
nor (depending on the installed font and the result he gets).

PR #1312, opened 2021-05-24
Debug update

orazioedoardo:
- Updated the FAQ link to point to [docs.pivpn.io](docs.pivpn.io).
- Tweaked the output header such that it doesn't show the full commit message, just the subject. After all we only need the git information to know when the user has installed PiVPN.
- Added the branch to the output.


4s3ti:
LGTM feel free to go ahead.

PR #1313, opened 2021-05-24
Remove kernel version check on Raspbian

orazioedoardo:
Raspbian images have been shipping with kernel 5.10.17 since March 4. This kernel version has the WireGuard module built in which means we don't need to check if the user has updated, neither to install the module via DKMS.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1242,53 +1242,6 @@ installWireGuard(){
 
 	if [ "$PLAT" = "Raspbian" ]; then
 
-		# If the running kernel is older than the kernel from the repo, dkms will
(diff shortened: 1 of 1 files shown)


4s3ti:
hmm .. What about users that might be using older versions?

orazioedoardo:
> hmm .. What about users that might be using older versions?

If the user has flashed Raspbian after March 4 or if the user has flashed an earlier version of Raspbian but has upgraded packages  at least in mid January (when kernel 5.10 was released in APT repos), his kernel version is compatible. I think it's reasonable to assume it did.

4s3ti:
Sounds good to me! Merging then.

PR #1321, opened 2021-06-05
Add wireguard persistentkeepalive

DerDanilo:
Added optional PERSISTENTKEEPALIVE var for wireguard client profiles. 

Only added when defined in `setupVars.conf `for unattended installation or when added manually to `/etc/pivpn/wireguard/setupVars.conf` post server setup to be used during client profile generation.


4s3ti (changes requested):
Thanks a lot!

Can you also please put up a MR to this: https://github.com/pivpn/docs/blob/master/docs/wireguard.md  with the documentation about it?

README.md
```diff
@@ -108,7 +108,7 @@ Please be respectful and be aware that this is maintained with our free time!
 
 for community support or general questions.
 Feel free to post on our subreddit <https://www.reddit.com/r/pivpn/>
-You can also join #pivpn <ircs://freenode/pivpn> on freenode in IRC
+You can also join #pivpn on [libera.chat](https://libera.chat) IRC network
```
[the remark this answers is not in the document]
DerDanilo: Yeah maybe. I merged the latest master state before creating a branch. The branch should have been clean since I commited the change a few hours later.

DerDanilo:
> Thanks a lot!
> 
> Can you also please put up a MR to this: https://github.com/pivpn/docs/blob/master/docs/wireguard.md with the documentation about it?

Will do, but let's keep this separate for now to get this option upstream.

4s3ti:
I does look good to me, however ill wait for @orazioedoardo to have some time to look at this and provide some input.

4s3ti (approved):

examples/unattended_wireguard_example.conf
```diff
@@ -9,4 +9,5 @@ pivpnPORT=51820
 pivpnDNS1=9.9.9.9
 pivpnDNS2=149.112.112.112
 pivpnHOST=pivpn.example.com
+pivpnPERSISTENTKEEPALIVE=21
```
orazioedoardo, line 12: WireGuard website suggests 25 as default value: https://www.wireguard.com/quickstart/#nat-and-firewall-traversal-persistence

DerDanilo: That might be and I also update to the suggested value but 21 is the value that we found working better than the 25 seconds. However I understand that we should probably go for the suggested value.

DerDanilo:
Please merge :)

Issue #1282, opened 2021-03-21
[Support] PiVPN installation prevents pi-hole from being reachable from non-VPN clients.


Ge0rges:
## In raising this issue, I confirm the following:

- [X] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [] The issue I am reporting can be *replicated*.
- [] The issue I am reporting can be *is* directly related to the pivpn installer script.
- [X] The issue I am reporting isn't a duplicate 

### Issue
Hello all, I had a working setup running PiHole and Unbound as the upstream DNS. I had UFW set up, such that only my home IP could make requests to PiHole (ie. only my home IP packets will be accepted on port 53). This setup worked. 

After installing PiVPN (Wireguard), devices on my home network can no longer resolve DNS queries. This is confirmed by the PiHole log not showing any queries being requested. TCPDump shows that these packets are getting to the server however. 

Unbound listens on 5353. PiHole listens on 53. Apache is also running on this server on 80 and 443. I use UFW as my firewall.

### Have you searched for similar issues and solutions?
https://github.com/pivpn/pivpn/issues/1207
https://github.com/pivpn/pivpn/issues/380
https://github.com/pivpn/pivpn/issues/1192

### Console output of      `curl -L install.pivpn.io | bash`
```
root@ge0rges:~# curl -L install.pivpn.io | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    17  100    17    0     0    202      0 --:--:-- --:--:-- --:--:--   202
100   162  100   162    0     0    830      0 --:--:-- --:--:-- --:--:--   830
100 83358  100 83358    0     0   193k      0 --:--:-- --:--:-- --:--:--  193k
:::
::: You are root.
::: Update option selected.
::: The updating functionality for PiVPN scripts is temporarily disabled
::: To keep the VPN (and the system) up to date, use 'apt update' and 'apt upgrade'
```

### Console output of      `pivpn add` or `pivpn add nopass`
```
  Enter a Name for the Client: foo
::: Client Keys generated
::: Client config generated
::: Updated server config
::: Updated hosts file for Pi-hole
::: WireGuard reloaded
======================================================================
::: Done! foo.conf successfully created!
::: foo.conf was copied to /home/pivpn/configs for easy transfer.
::: Please use this profile only on one device and create additional
[...]

PR #1410, opened 2021-11-18
Allow DNS incoming requests through UFW

EWouters:
Fixes https://github.com/pivpn/pivpn/issues/1282

Also added a line to the uninstall script to remove the rule at uninstall time.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1473,6 +1473,12 @@ askClientDNS(){
 
 			echo "pivpnDNS1=${pivpnDNS1}" >> ${tempsetupVarsFile}
 			echo "pivpnDNS2=${pivpnDNS2}" >> ${tempsetupVarsFile}
+
+			# Allow DNS incomming requests through UFW.
+			if [ "$USING_UFW" -eq 1 ]; then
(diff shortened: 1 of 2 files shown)


4s3ti:
lgtm, Thanks a lot! 

Merging. 
I might merge test into master this weekend.

auto_install/install.sh
```diff
@@ -1473,6 +1473,12 @@ askClientDNS(){
 
 			echo "pivpnDNS1=${pivpnDNS1}" >> ${tempsetupVarsFile}
 			echo "pivpnDNS2=${pivpnDNS2}" >> ${tempsetupVarsFile}
+
+			# Allow DNS incomming requests through UFW.
```
EWouters, line 1477: I messed up the spelling of the comment, I spelled incoming with two m's. I thought I fixed it but it is still written with two m's.

4s3ti: hmm ... Nope its actually correct ... https://github.com/pivpn/pivpn/blob/test/auto_install/install.sh#L1477

That is probably older version

EWouters: Strange, but fine :)

orazioedoardo: @4s3ti @EWouters I’ve changed it, that this conversation is marked as resolved.

PR #1426, opened 2021-12-07
Fix typo in removeOVPN.sh for default action

nilsknacken:
The action if the user does not input anything when revoking a certificate is to not revoke. Let the text reflect that by making N uppercase.


4s3ti:
LGTM

PR #1455, opened 2022-02-11
Basic wireguard IPv6 (NAT) support

DerDanilo:
Following my comment (https://github.com/pivpn/pivpn/discussions/1394#discussioncomment-2159928) I tried to integrate IPv6 for pivpn.

---

This PR adds basic IPv6 support for Wireguard using IPv6 NAT. IPv6 NAT is more or less ugly but works fine, since requests are desired  be coming from the public IPv6 of the VPN server and not the client IPv6 (in most cases). 

- Since the IPv6 interface can differ from the IPv4 interface it's configurable. Usually it's the same.
- I am not sure what has to be done to have all tests run successfully since IPv6 has `:` and `::` which might cause some trouble.
- I have no experience with CLI GUI stuff, hence I hope someone can adjust the code if it's not working

WG IPv6 basically works on multiple installations I am maintaining. Now some people need to test and help finalize it to make this work with pivpn.

Edit: I fixed lint errors but don't know what is wrong with the setup tests since the setup seems to finish without errors but wireguard doesn't seem to be able to start. Can someone please check on that?

Thanks!


DerDanilo:
> Tried it out, log said:
> 
> ```
> ip6tables v1.8.2 (legacy): Empty interface is likely to be undesired
> Try `ip6tables -h' or 'ip6tables --help' for more information.
> ```
> 
> I can't get it to connect. Also, shouldn't there be an IPv6 subnet in the `Address` field in the configs like [this](https://github.com/pivpn/pivpn/discussions/1394#discussioncomment-1570174)?
> 

Thanks for that info! 
Reading your post-error comments was fun! IPv6 can be annoying sometimes but it's actually not much different than IPv4.

4s3ti:
@DerDanilo not sure if you are local testing before hand, but just want to make sure you  please do try some local testing before pushing to the MR, Travis CI has usage limits and even tho they are quite kind in raising them if needed i'd rather avoid it. 


Once all tests are passing and you feel code is ready to go let me know. 
I will try to find some time during the week to assist, review and test.

DerDanilo:
@4s3ti I just launched a clean test machine today to do more testing even though I thought I tested everything already. Found the bug with the missing IPv6dev reference, which is fixed now.

The annoying thing with local tests is that the install script will always install the master upstream repo (into pivpnFilesDir="/usr/local/src/pivpn"), which can be changed by hand of course, but the branch then is wrong, since I am not pushing against the fork master obviously. Maybe this can be improved for future testing, allowing providing such option on the CLI via options or maybe just env vars. 'custom_git_repo_url' and 'custom_git_branch' (to checkout after clone).
These options then should also be referenced in the tests to have those test properly. I don't have much experience with Travis and hence don't know how that behaves. 

Looking forward to have this fixed. Any help is appreciated.

---

Other question: Can we add an option to disable any script interaction with the firewall at all? I understand that this should be handled by the script for folks that don't work on it otherwise. But it is a valid option to have the script not fiddle with any firewall if the user wishes so. Could be an option in the unattended file and should possible be an additional question in the whiptail dialog.

4s3ti:
I am really struggling to find time :(   

@DerDanilo there are ways to test the pipelines locally .. but to be honest .. the easy way is really to just push and look at the build :p 

What i usually do is to spin up a cloud instance and try to replicate the build environment and see what happens. 

I usually use hetzner as my cloud provider for this kind of things, and i've noticed it gets quite close to the  build environment, however, but i believe any clean debian system might work.

orazioedoardo:
> There is also the issue with iptables being replace by nftables. We could also add support for that.

I'm planning to do this.

> Why? Most people don't fiddle with IPv6 on their system, hence link local addresses are enabled by default.

I mean actual IPv6 internet connectivity, a global unicast address and possibly a reachability test.

> I think that it should be clear that any firewall in front of the server needs editing to make this work.

That comment was a suggestion for the user above, clearly you can't change the router firewall from the Pi, unless UPnP is enabled (I think automatic port forwarding was requested in the past).

> It is also not required if one wishes to connect to a DNS name which is the best option here. Those can contain dynamic IPs without any config change requirements. With dual stack a DNS name is anyways the best option, if one wishes to use that. So people either put their IP there or put a DNS name.

Though currently the whiptail prompt has options for the IPv4 address or a domain name. **In case IPv6 is supported**, there are some options:

1) Show both IPv4 and IPv6 address, and create two configurations if IP is selected instead of the DNS name.
2) Only provide the option for DNS same and say why it is the best option.

>  For the sake of dynamic IPv6 addresses (yes some providers even rotate IPv6 subnets) and that most pis have the 'privacy/private' IPv6 option enabled I skipped that.

Ok, speaking of this, I think the script should ask to set a static IPv6 address too.

DerDanilo:
> > There is also the issue with iptables being replace by nftables. We could also add support for that.
> 
> I'm planning to do this.

Would you mind including an option to not touch any firewall rules at all in the same PR?
Should at least be in the unattended file but maybe also in the whiptail commands.
Something like "it looks like you're using iptables/ufw/nftables, should pivpn configure your firewall for wireguard?"

> > Why? Most people don't fiddle with IPv6 on their system, hence link local addresses are enabled by default.
> 
> I mean actual IPv6 internet connectivity, a global unicast address and possibly a reachability test.

Maybe we can just do a `ping6 google.com` or something like this. If this works out IPv6 should be fine. What do you think?
Any suggestion how to implement that?
We should maybe also provide an option that disables any IP checks in case of problems there.

> > I think that it should be clear that any firewall in front of the server needs editing to make this work.
> 
> That comment was a suggestion for the user above, clearly you can't change the router firewall from the Pi, unless UPnP is enabled (I think automatic port forwarding was requested in the past).

UPnP is not reliable enough I think. But yes, it could also be an option to add support for that. Please don't make this the default.

I suggest you also include this feature in your firewall PR (nftables).

> > It is also not required if one wishes to connect to a DNS name which is the best option here. Those can contain dynamic IPs without any config change requirements. With dual stack a DNS name is anyways the best option, if one wishes to use that. So people either put their IP there or put a DNS name.
> 
> Though currently the whiptail prompt has options for the IPv4 address or a domain name. **In case IPv6 is supported**, there are some options:
> 
>     1. Show both IPv4 and IPv6 address, and create two configurations if IP is selected instead of the DNS name.
> 
>     2. Only provide the option for DNS same and say why it is the best option.

Why put a static IP into the config? All configs have to be updated if any IP was changed. 
Are you suggesting to create two different config files per user if the user chooses to use both IP stacks instead of a DNS name? 
I suggest to simply say that a user may use a single IP as Endpoint or a DNS based endpoint to support dual stack.

From some reddit post:
> Wireguard for Android/iOS will prefer IPv4 if the endpoint hostname you specified successfully resolves AAAA and A, to help  with roaming between potentially IPv4-only networks. You can enforce IPv6 as transport by using a AAAA-only endpoint. wg-quick will prefer IPv6 if possible.

AFAIK pivpn doesn't yet support to properly update existing profiles. I am also missing an option to override the "Allowed IPs" var per user. Yes users may simply add the nets their routing through the VPN but we have some cases where changing the list on a per user basis is very useful and avoids editing the config afterwards.

> > For the sake of dynamic IPv6 addresses (yes some providers even rotate IPv6 subnets) and that most pis have the 'privacy/private' IPv6 option enabled I skipped that.
> 
> Ok, speaking of this, I think the script should ask to set a static IPv6 address too.

Are you referring to the part where the script will change the servers IP address? In general I think that this should be removed because the user should at least understand how to configure an IP address when going for a VPN.
As said; IPv6 on common internet providers are also dynamic for privacy reasons. Hence that should not be the default at least.

orazioedoardo:
> Would you mind including an option to not touch any firewall rules at all in the same PR?

Ok.

> Maybe we can just do a ping6 google.com or something like this. If this works out IPv6 should be fine. What do you think?

Should be enough according to my limited knowledge of IPv6.

> We should maybe also provide an option that disables any IP checks in case of problems there.

Which problems could arise by checking for connectivity?

> UPnP is not reliable enough I think. But yes, it could also be an option to add support for that. Please don't make this the default.

Definitely not default. UPnP is often seen as a security risk so I'm not sure if we want to support such feature, although it would simplify the work for the user.

> Are you suggesting to create two different config files per user if the user chooses to use both IP stacks instead of a DNS name?

Yes, but see below too.

> AFAIK pivpn doesn't yet support to properly update existing profiles.

Yes, this is quite implicitly requested.

> I suggest to simply say that a user may use a single IP as Endpoint or a DNS based endpoint to support dual stack.

Which one? IPv4, IPv6, or ask the user?

> Are you referring to the part where the script will change the servers IP address? In general I think that this should be removed because the user should at least understand how to configure an IP address when going for a VPN.

I'm referring to the part of the script that sets the static LAN IP. In the same way, I think it should set a static IPv6 but warn that some ISPs change the prefix from time to time (even if is not recommended by RIPE: https://www.ripe.net/publications/docs/ripe-690). If we are removing tasks that the script has a chance to perform, then the script itself is less useful to the user.

orazioedoardo:
> External firewall rules, simply not required because the admin doesn't want any checks. DNS names being blocked. etc.

@4s3ti can you add an IPv6 address to `install.pivpn.io`?

> Let the user choose, so ask. Suggest IPv4 if in doubt because of the mentioned mobile restrictions.

ACK

> If you want to spend time for that okay fine with me. Just please don't let this PR stale forever because nobody finds time to do all of the suggested. We should sort the suggested and split parts into different PRs.

Separate pull requests. I will test this as soon as i have time IRL, and find out why the checks fail.

DerDanilo:
> @4s3ti can you add an IPv6 address to `install.pivpn.io`?

I would advise against using such subdomain for that. Rather create a specific domain for that purpose.
I can provide simple PHP scripts to provide the public IP (v4 or v6) of the device connecting to the subdomain.
I suggest: ipv4.pivpn.io and ipv6.pivpn.io for that.
This way we'd also have the public IP and can asure that this will work for the installation script. Doesn't really require anything. Just PHP on the server.

> > If you want to spend time for that okay fine with me. Just please don't let this PR stale forever because nobody finds time to do all of the suggested. We should sort the suggested and split parts into different PRs.
> 
> Separate pull requests. I will test this as soon as i have time IRL, and find out why the checks fail.

I am happy to further help improve pivpn but I rather go for smaller packages adding improvements than one or two huge PRs.

@4s3ti Can we maybe discuss out of public Github regarding the dev env and testing? Maybe a discord group or something that also allows voice. Maybe you can create one and provide the info in the repo README ("for devs only, no support!" with manual server approval from your side).

4s3ti:
> @4s3ti can you add an IPv6 address to install.pivpn.io?

Not sure ... need to investigate that. 

> @4s3ti Can we maybe discuss out of public Github regarding the dev env and testing? Maybe a discord group or something that also allows voice. Maybe you can create one and provide the info in the repo README ("for devs only, no support!" with manual server approval from your side).

We have matrix where we can chat .. also can make a dev only room on matrix .. would that suffice?

DerDanilo:
> We have matrix where we can chat .. also can make a dev only room on matrix .. would that suffice?

Where can I sign up? 

> > @4s3ti can you add an IPv6 address to install.pivpn.io?
> 
> Not sure ... need to investigate that.

Hetzner cloud for a few bugs is available with both.

4s3ti:
> Hetzner cloud for a few bugs is available with both. 

I know .. but I have everything running on Google Kubernetes service, so configuring IPv6 might require some tricks .. need to investigate that

As for matrix .. the easiest place is here: https://element.io/  the only reason why i am not opening up a discord is because it becomes way to much stuff to manage. the room is on the README ... and during the weekend can spin up a devs room or something. will think the best approach.

orazioedoardo:
> I would advise against using such subdomain for that. Rather create a specific domain for that purpose.

The rationale behind `install.pivpn.io` is that the user has likely visited the website so it is implicitly trusted and available.

DerDanilo:
lets discuss further in matrix. To much discussion here.

DerDanilo:
> > I would advise against using such subdomain for that. Rather create a specific domain for that purpose.
> 
> The rationale behind `install.pivpn.io` is that the user has likely visited the website so it is implicitly trusted and available when the script is being run.

We might switch to that URL in the future. For the time being I implemented just testing at google.com as this is most probably also allowed and google.com supports IPv6 of course.

DerDanilo:
After discussing with @4s3ti in matrix we decided to disable IPv6 tests in travis since travis doesn't really support IPv6 properly. It's only available in the LXC container which are only running on ARM which brings other problems to the table.

I implemented the following:

- A switch to disable IPv6 config
- IPv6 is enabled by default IF
--> IPv6 simple ping test, that will automatically disable ipv6 if not working (using google.com for now, may be switched to install.pivpn.io in future, no IPv6 address at the moment)
- allow disabling IPv6 via unattended file switch (mainly for testing exclusion, since ipv6 should be used if available)

We need someone to test this now without and with unattended file. Disabling IPv6 via unattended file works already otherwise travis tests would fail.

PR #1467, opened 2022-03-13
Runtime network configuration

orazioedoardo:
This pull requests changes the way firewall configuration and sysctl settings are applied. Instead of statically modifying sysct/ufw/iptables files, a systemd [override](https://github.com/orazioedoardo/pivpn/blob/bc0a97fac9f6a05329fba9096270d1e66ffa0d8d/files/etc/systemd/system/wg-quick@wg0.service.d/pre-post-jobs.conf) is added to the wireguard and openvpn units to change the settings before starting the vpn and reset them after the vpn is stopped.

Benefits:
1) Files are not modified on disk (except ufw), changes are made when they are needed.
2) Trivial to change the physical network interface, simply run `/opt/pivpn/network/networkCONF.sh wireguard reset`, edit `/etc/pivpn/wireguard/setupVars.conf`, run `/opt/pivpn/network/networkCONF.sh wireguard apply`.
3) Easy to give the ability to skip setting up the firewall on installation (@DerDanilo asked for this).
4) [Only one place](https://github.com/orazioedoardo/pivpn/blob/bc0a97fac9f6a05329fba9096270d1e66ffa0d8d/scripts/network/networkCONF.sh#L17-L41) where the firewall definitions are saved, instead of install, self check, uninstall.


scripts/network/networkCONF.sh
```diff
@@ -0,0 +1,168 @@
...
+	exit 1
+fi
+
+# shellcheck disable=SC1090
+source "${setupVars}"
+
+scriptDir="/opt/pivpn"
+
+# shellcheck disable=SC2034
+loadFirewallRulesArrays(){
```
DerDanilo, line 17: I know it's still WIP but I think that this section should check if IPv6 is enabled or disabled. E.g. a IPv6 device is not defined when the IPv6 ping check fails. Hence some of the following commands will fail.

DerDanilo:
This PR brings a lot of changes but I like that firewall and sysctl settings are done in a cleaner way. Waiting until you're ready to merge and I'll try to find time checking them.

Is nftables still on the todo list for this PR?

scripts/network/networkCONF.sh
orazioedoardo: Thanks, I will fix this. It doesn't actually cause problems because all those boring functions in the `*FUNC.sh` files do check for IPv6, but yes, some of these arrays `loadFirewallRulesArrays` have undefined variables now.

orazioedoardo:
> Is nftables still on the todo list for this PR?

No. I want to take more time and figure out something that can enable [this functionality too](https://github.com/pivpn/pivpn/discussions/1374).

orazioedoardo:
So this pull request went stale because I wasn't really satisfied with the changes. Too many overhead scripts and indirection. Would like to move to nftables but it's not easy to programmatically manipulate like iptables unless it's a clean install.

PR #1470, opened 2022-03-13
Fix potential IPv6 leak, add IPv6 options

DerDanilo:
- add option to force IPv6 config
- forced IPv6 route when server has no IPv6 uplink to prevent IPv6 leak
- add IPv6 options for cli and unattended file

First time working with wiptale. Can someone check please? It should default to "yes" (forcing IPv6 route).


auto_install/install.sh
```diff
@@ -753,6 +783,24 @@ checkipv6uplink(){
...
+askforcedipv6route(){
+	if [ "${runUnattended}" = 'true' ]; then
+		echo "::: Enable forced IPv6 route with no IPv6 uplink on server."
+		echo "pivpnforceipv6route=${pivpnforceipv6route}" >> ${tempsetupVarsFile}
+		return
+	fi
+
+	whiptail --msgbox --backtitle "Privacy setting" --title "IPv6 leak" "Since this server doesn't seem to have a working IPv6 connection or IPv6 was disabled on purpose, it is still recommended you force all IPv6 connections through the vpn.\\nThis will prevent an IPv6 client leak.\\nIt might cause the client to have slow reponse when browsing the web though.\\nDisable when you having trouble with this." ${r} ${c}
+
+	if (whiptail --backtitle "Privacy setting" --title "IPv6 leak" --yesno "Do you want to disable forced IPv6 route?" ${r} ${c}); then
```
orazioedoardo, line 795: I would use a single `--yesno` whiptail dialog with this text:

```
Although this server doesn't seem to have a working IPv6 connection or IPv6 was disabled on purpose, it is still recommended you force all IPv6 connections through the VPN.\\n\\nThis will prevent the client from bypassing the tunnel and leaking its real IPv6 address to servers, though it might cause the client to have slow response when browsing the web on IPv6 networks.\\n\\nDo you want to force routing IPv6 to block the leakage?
```

> Disable when you having trouble with this.

I've removed this because after install you can't "officially" change it when creating a client.

DerDanilo: thanks, done

PR #1433, opened 2021-12-25
Randomize wg internal subnet address

orazioedoardo:
Satisfies feature request https://github.com/pivpn/pivpn/discussions/1424.

Issue #1267, opened 2021-03-05
[BUG] Improper IP Address after subnetclass change


Matt-L-KZ7MFL:
## In raising this issue, I confirm the following: 

`{please fill the checkboxes, e.g: [X]}`

- [x] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*.
- [x] The issue I am reporting *is* directly related to the pivpn installer script.
- [x] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pivpn/pivpn/wiki/FAQ), [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen)).

**Describe the bug**
After changing the pivpnNET to 10.0.5.16 and subnetClass to 28 it should produce IP's between 10.0.5.17-30. Instead it produces an improper IP of 10.0.5..2/28. I have also changed the wg0.conf file to 10.0.5.16/28.

**To Reproduce**
1. Install PiVPN normally using Pi-Hole and WireGuard.
2. Edit the wg0.conf file to address = 10.0.5.16/28
3. Edit setupVars.conf to match the included file.
4. Restarted system.
5. ran pivpn -a named TEST.
6. Check /home/pi/configs/TEST.conf for IP address.

Side Note: after changing numerous files that record the client configured IP to 10.0.5.17 I can prove that PiVPN/WireGuard work as intended. 

EDIT: I should mention i'm trying to get piVPN to work in the 10.0.5.16/28 subnet which has addresses from 10.0.5.16-30. Not the 10.0.5.0/28 subnet which has addresses from 10.0.5.0-15.

orazioedoardo:
`pivpnNET` should be `10.0.5.0` with 0 at as the least significant octet. Also you need to update `/etc/iptables/rules.v4`

orazioedoardo:
I've removed your log because it contained an actual configuration. You can send `pivpn debug` as text instead of attachment.

Matt-L-KZ7MFL:
> `pivpnNET` should be `10.0.5.0` with 0 at as the least significant octet. Also you need to update `/etc/iptables/rules.v4`

[...]

PR #1505, opened 2022-04-21
Give user option to choose their network for setupVars file

Normanras:
I have multiple Wireguard networks in my house, used for home lab stuff along with various family members who need certain permissions. After installing PiVPN and using the default network (10.6.0.0) and then changing it later in the various files in /etc/pivpn etc, I kept getting network errors. However, by changing and choosing the network from the very beginning with your install script, none of those errors occurred. 

I thought about making the warning more explicit and/or steering most users to using the default network, so I'm open to suggestions and changes. Very few people will actually use it, but having the flexibility is nice. 

Also... this is my first PR so go easy. :) Great work on this project, though! I really love PiVPN.


auto_install/install.sh
```diff
@@ -1087,10 +1087,22 @@ setVPNDefaultVars(){
...
+					pivpnNET=$answer
+				else
+					whiptail --msgbox --backtitle "Setup PiVPN" --title "Installation Mode" "$answer is not a valid IP. Please try again." 8 78
+					continue
+				fi
+			else
+				whiptail --msgbox --title "DEFAULT" "Great! Using the default IP of 10.8.0.0" 8 78
+				pivpnNET="10.8.0.0"
+			fi
+		done
```
4s3ti, line 1105: I think this solution breaks unattended setup, it does not seem to assume a default in case nothing is provided (when running unattended)

auto_install/install.sh
```diff
@@ -1099,10 +1111,22 @@ setWireguardDefaultVars(){
...
+					pivpnNET=$answer
+				else
+					whiptail --msgbox --backtitle "Setup PiVPN" --title "Installation Mode" "$answer is not a valid IP. Please try again." 8 78
+					continue
+				fi
+			else
+				whiptail --msgbox --title "DEFAULT" "Great! Using the default IP of 10.6.0.0" 8 78
+				pivpnNET="10.6.0.0"
+			fi
+		done
```
4s3ti, line 1129: I think this solution breaks unattended setup, it does not seem to assume a default in case nothing is provided (when running unattended)

4s3ti (commented):
Hi, thanks a lot for your PR, and for trying to contribute. 

Being this your first one, let me give you a couple of general notes: when contributing to any opensource project its a good idea to read their documentation and try to find if there are any notes about contributions and how do do so. 

For pivpn here's what you should know: https://github.com/pivpn/pivpn#general-guidelines
And here: https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md

Also, what you are trying to implement is already supported via unattended installation: https://docs.pivpn.io/install/#non-interactive-installation

We have decided to leave advanced features out of whiptail dialogs for multiple reasons, one of them is to not overwhelm less technically savvy users with information therefore causing more errors / issues during installation. 

More technically savy users are assumed to be more inclined to read documentation and therefore understand what they are doing so unattended install is the way to override some defaults and get a bit more technical. 

Sometimes i understand that documentation might not be very updated or accurate, but in most cases (pivpn included) documentation is also opensource and can be also improved by everybody. 

pivpn case, documentation source is here: https://github.com/pivpn/docs

This being said, I don't think this PR should be approved, for multiple reasons, some of them mentioned in this review.  however I am pinging @orazioedoardo @TinCanTech which usually have great feedback to provide, and may correct me here if i am wrong.

orazioedoardo:
There was a [feature request](https://github.com/pivpn/pivpn/discussions/1424) and subsequent [commits](https://github.com/pivpn/pivpn/pull/1433/files) to randomize the virtual network range, so this pull request kinda overlaps, really.

> We have decided to leave advanced features out of whiptail dialogs for multiple reasons, one of them is to not overwhelm less technically savvy users with information therefore causing more errors / issues during installation.

I agree with @4s3ti here.

orazioedoardo:
> I opened this mainly in reply to https://github.com/pivpn/pivpn/issues/1267. So given your above comment, can 1267 be closed then? Especially if you're staying away from anything advanced in whiptail?

If the problem you are trying to solve is "missing ability to specify a custom subnet", then you can [already do that](https://github.com/pivpn/pivpn/pull/1279) via unattended configuration. On the other hand if the problem is "10.6.0.0/24" overlaps with other range, this is already fixed by [randomizing the range](https://github.com/pivpn/pivpn/pull/1433). I think it is useful to have the ability to set a custom subnet but less tech savvy probably won't do it, and more experienced uses can use the unattended setup. 

Issue #1267 is about changing the subnet after the fact, not at install time. Pull request #1467 will make that easier to change the subnet by simply editing `/etc/pivpn/.../setupVars.conf` although it doesn't fix the issue specifically as the script [still assumes /24](https://github.com/pivpn/pivpn/blob/c7bbb9fda21d641071e58145a2d51766bea33082/scripts/wireguard/makeCONF.sh#L99-L106), so the issue can't be closed.

Normanras:
Ah, okay, got it. I see the comment now asking about accounting for different subnet masks. Closing this and I'll open a new PR if I get it done according to #1267. Thanks for the feedback.

PR #1506, opened 2022-04-23
Update install.sh

Sudo-null7:
Just bugs me

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -915,7 +915,7 @@ chooseUser(){
 	fi
 
 	# Explain the local user
-	whiptail --msgbox --backtitle "Parsing User List" --title "Local Users" "Choose a local user that will hold your ovpn configurations." ${r} ${c}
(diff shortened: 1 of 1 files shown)


4s3ti (changes requested):
https://github.com/pivpn/pivpn#feedback--suppor

> Please carefully read the [contributors' guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md)

https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md#do-you-want-to-open-a-pull-request-to-modify-the-source

> Open a new GitHub pull request with the patch on the [test](https://github.com/pivpn/pivpn/tree/test) branch.
>
> Ensure the PR description clearly describes the problem and solution. Include the relevant issue number if applicable.

Now with "bureaucracy"  out of the way ... lets talk about the change itself. 

Back in the days, would make sense to call it  "ovpn configurations" because pivpn was only installing OpenVPN .. but now since we are also installing wireguard, clearing that phrasing might be wise. 

*s/account/client/*:  Change account to client, there is no accounts with pivpn, rather client configuration files.

4s3ti:
Also bugs me the amount of time with no answer after changes have been suggested.

PR #1572, opened 2022-07-21
Reformatted the code

giulio-coa:
I did some style improvements

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,11 @@ Please report unacceptable behavior to any project maintainer.
 
 * Ensure the PR description clearly describes the problem and solution. Include the relevant issue number if applicable.
 
+* Use the following [commit rules](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format).
+
+* Use the following [code style rules](https://google.github.io/styleguide/shellguide.html).
+We suggest you to use `shfmt` with the options `-i 2 -ci -sr -w`
+
 ### **PiVPN Website**
 
 * Yes the site is open source as well! You can find and contribute to [pivpn.io](https://github.com/pivpn/pivpn.io) directly.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,7 +25,7 @@ Please report unacceptable behavior to any project maintainer.
 * Use the following [commit rules](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format).
 
 * Use the following [code style rules](https://google.github.io/styleguide/shellguide.html).
-We suggest you to use `shfmt` with the options `-i 2 -ci -sr -w`
+We suggest you to use `shfmt` with the options `-i 2 -ci -sr -w -bn`
 
 ### **PiVPN Website**
 
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -661,30 +661,37 @@ preconfigurePackages() {
     # to install wireguard-tools.
     [[ "${WIREGUARD_BUILTIN}" -eq 1 && -n "${AVAILABLE_WIREGUARD}" ]] \
       ||
+      
       # If the package is not available, on Debian and Raspbian we can
       # add it via Bullseye repository.
       [[ "${WIREGUARD_BUILTIN}" -eq 1 && ("${PLAT}" == 'Debian' || "${PLAT}" == 'Raspbian') ]] \
       ||
+      
       # If the module is not builtin, on Raspbian we know the headers
       # package: raspberrypi-kernel-headers
       [[ "${PLAT}" == 'Raspbian' ]] \
       ||
+      
       # On Alpine, the kernel must be linux-lts or linux-virt if we want to
       # load the kernel module
       [[ "${PLAT}" == 'Alpine' && ! -f /.dockerenv && "$(uname -mrs)" =~ ^Linux\ +[0-9\.\-]+\-((lts)|(virt))\ +.*$ ]] \
       ||
+      
       # On Alpine Docker Container, the responsibility to have a WireGuard
       # module on the host system is at user side
       [[ "${PLAT}" == 'Alpine' && -f /.dockerenv ]] \
       ||
+      
       # On Debian (and Ubuntu), we can only reliably assume the headers package
       # for amd64: linux-image-amd64
       [[ "${PLAT}" == 'Debian' && "${DPKG_ARCH}" == 'amd64' ]] \
       ||
+      
       # On Ubuntu, additionally the WireGuard package needs to be available,
       # since we didn't test mixing Ubuntu repositories.
       [[ "${PLAT}" == 'Ubuntu' && "${DPKG_ARCH}" == 'amd64' && -n "${AVAILABLE_WIREGUARD}" ]] \
       ||
+      
       # Ubuntu focal has wireguard support
       [[ "${PLAT}" == 'Ubuntu' && "${DPKG_ARCH}" == 'arm64' && "${OSCN}" == 'focal' && -n "${AVAILABLE_WIREGUARD}" ]]
   then
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -139,13 +139,13 @@ main() {
     echo "::: Forced IPv6 config, skipping IPv6 uplink check!"
     pivpnenableipv6=1
   else
-    if [[ -z "${pivpnenableipv6}" ]] ||
-      [[ "${pivpnenableipv6}" -eq 1 ]]; then
+    if [[ -z "${pivpnenableipv6}" ]] \
+      || [[ "${pivpnenableipv6}" -eq 1 ]]; then
       checkipv6uplink
     fi
 
-    if [[ "${pivpnenableipv6}" -eq 0 ]] &&
-      [[ "${pivpnforceipv6route}" -eq 1 ]]; then
+    if [[ "${pivpnenableipv6}" -eq 0 ]] \
+      && [[ "${pivpnforceipv6route}" -eq 1 ]]; then
       askforcedipv6route
     fi
   fi
@@ -155,8 +155,8 @@ main() {
   if checkStaticIpSupported; then
     getStaticIPv4Settings
 
-    if [[ -z "${dhcpReserv}" ]] ||
-      [[ "${dhcpReserv}" -ne 1 ]]; then
+    if [[ -z "${dhcpReserv}" ]] \
+      || [[ "${dhcpReserv}" -ne 1 ]]; then
       setStaticIPv4
     fi
   else
@@ -294,8 +294,8 @@ checkExistingInstall() {
     fi
   fi
 
-  if [[ -z "${UpdateCmd}" ]] ||
-    [[ "${UpdateCmd}" == "Reconfigure" ]]; then
+  if [[ -z "${UpdateCmd}" ]] \
+    || [[ "${UpdateCmd}" == "Reconfigure" ]]; then
     :
   elif [[ "${UpdateCmd}" == "Update" ]]; then
     ${SUDO} "${pivpnScriptDir}/update.sh" "$@"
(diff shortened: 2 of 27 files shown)


orazioedoardo:
Thanks but I think we should convert the entire script to a code style we agree on and then start requiring it, instead of spot pull requests. I'm about to start a discussion now.

auto_install/install.sh
```diff
@@ -2411,22 +2412,26 @@ confLogging() {
...
+		echo $'\t' 'missingok'
+		echo $'\t' 'notifempty'
+		echo $'\t' 'compress'
+		echo $'\t' 'delaycompress'
+		echo $'\t' 'sharedscripts'
+		echo $'\t' 'postrotate'
+		echo $'\t' $'\t' 'invoke-rc.d rsyslog rotate >/dev/null 2>&1 || true'
+		echo $'\t' 'endscript'
+		echo '}'
+	} | $SUDO tee /etc/logrotate.d/openvpn > /dev/null
```
orazioedoardo, line 2434: I'm not a fan of this change anyway, I think the previous was more readable.

orazioedoardo (commented):
Considering discussion https://github.com/pivpn/pivpn/discussions/1573 and my previous [comment](https://github.com/pivpn/pivpn/pull/1572#issuecomment-1191499180), I wouldn't merge this pull request.

giulio-coa:
If we can determine the code style rules, this PR can become the one that will reformat the code

giulio-coa:
As mentioned in discussion #1573, I have reformatted the code according to the [Google rules](https://google.github.io/styleguide/shellguide.html)
Only file `auto_install/install.sh` is missing because I'm waiting for the merge of #1575

CONTRIBUTING.md
```diff
@@ -22,6 +22,11 @@ Please report unacceptable behavior to any project maintainer.
 
 * Ensure the PR description clearly describes the problem and solution. Include the relevant issue number if applicable.
 
+* Use the following [commit rules](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#-commit-message-format).
+
+* Use the following [code style rules](https://google.github.io/styleguide/shellguide.html).
+We suggest you to use `shfmt` with the options `--indent 2 --case-indent --space-redirects --write`
```
4s3ti, line 28:
```
$shfmt --help
usage: shfmt [flags] [path ...]

shfmt formats shell programs. If the only argument is a dash ('-') or no
arguments are given, standard input will be used. If a given path is a
directory, all shell scripts found under that directory will be used.

  -version  show version and exit

  -l        list files whose formatting differs from shfmt's
  -w        write result to file instead of stdout
  -d        error with a diff when the formatting differs
  -s        simplify the code
  -mn       minify the code to reduce its size (implies -s)

Parser options:

  -ln str        language variant to parse (bash/posix/mksh/bats, default "bash")
  -p             shorthand for -ln=posix
  -filename str  provide a name for the standard input file

Printer options:

  -i uint   indent: 0 for tabs (default), >0 for number of spaces
  -bn       binary ops like && and | may start a line
  -ci       switch cases will be indented
  -sr       redirect operators will be followed by a space
  -kp       keep column alignment paddings
  -fn       function opening braces are placed on a separate line

Utilities:

  -f        recursively find all shell files and print the paths
  -tojson   print syntax tree to stdout as a typed JSON

For more information, see 'man shfmt' and https://github.com/mvdan/sh.
```

options -i 2 -ci -sr  `shfmt -i 2 -ci -sr -w`

4s3ti (changes requested):
```
$find . -name *.sh -exec shfmt -l -i 2 -ci -sr {} \;
./auto_install/install.sh
```

this one is still missing, other than that all looks good.

giulio-coa:
The only rule that I have not followed to the letter is [this](https://google.github.io/styleguide/shellguide.html#s5.2-line-length-and-long-strings) since in some places it is not applicable or, if it is applied, it makes the code unreadable

4s3ti:
> The only rule that I have not followed to the letter is [this](https://google.github.io/styleguide/shellguide.html#s5.2-line-length-and-long-strings) since in some places it is not applicable or, if it is applied, it makes the code unreadable

can you provide and example? I am not sure if I am okay with that. if we are implementing rules for the code style then they should be followed.

giulio-coa:
> can you provide and example? I am not sure if I am okay with that. if we are implementing rules for the code style then they should be followed.

Some examples are: 
* https://github.com/giulio-coa/pivpn/blob/82a7cbfc27aab7badf3675ed9210d64e50c8c0a2/auto_install/install.sh#L648
* https://github.com/giulio-coa/pivpn/blob/82a7cbfc27aab7badf3675ed9210d64e50c8c0a2/auto_install/install.sh#L662-L682

4s3ti:
yeah those ones should be tricky ... its fine by me.

4s3ti (changes requested):
quick skim through, generally looks good, just a few comments/ changes.

ciscripts/startup.sh
```diff
@@ -1,47 +1,62 @@
...
 
-interface=$(ip -o link | awk '{print $2}' | cut -d':' -f1 | cut -d'@' -f1 | grep -v -w 'lo' | head -1)
-ipaddress=$(ip addr show "$interface" | grep -o -E "([0-9]{1,3}[\.]){3}[0-9]{1,3}/[0-9]{2}")
+interface=$(ip -o link |
+  awk '{print $2}' |
+  cut -d ':' -f 1 |
+  cut -d '@' -f 1 |
+  grep -v -w 'lo' |
+  head -1)
+ipaddress=$(ip addr show "${interface}" |
```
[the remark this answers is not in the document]
giulio-coa: In order to have less then 80 chars per line and a coherent code, I splitted it on every pipeline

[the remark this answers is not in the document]
giulio-coa: `shfmt` changed the formatting ...

4s3ti (approved):
LGTM

auto_install/install.sh
```diff
@@ -29,26 +28,34 @@ ovpnUserGroup="openvpn:openvpn"
...
 ######## PKG Vars ########
 PKG_MANAGER="apt-get"
-### FIXME: quoting UPDATE_PKG_CACHE and PKG_INSTALL hangs the script, shellcheck SC2086
+### FIXME: quoting UPDATE_PKG_CACHE and PKG_INSTALL hangs the script,
+### shellcheck SC2086
 UPDATE_PKG_CACHE="${PKG_MANAGER} update -y"
 PKG_INSTALL="${PKG_MANAGER} --yes --no-install-recommends install"
-PKG_COUNT="${PKG_MANAGER} -s -o Debug::NoLocking=true upgrade | grep -c ^Inst || true"
+PKG_COUNT="${PKG_MANAGER} -s -o Debug::NoLocking=true upgrade | "
+PKG_COUNT="${PKG_COUNT} grep -c ^Inst || true"
```
orazioedoardo, line 36: Changed to fit 80 character limit but I find this and similar multiline variable definitions confusing to read and too verbose.

auto_install/install.sh
```diff
@@ -29,26 +28,34 @@ ovpnUserGroup="openvpn:openvpn"
...
+# grep here. This way when uninstalling PiVPN we won't prompt to remove packages
+# that may have been installed by the user for other reasons
 INSTALLED_PACKAGES=()
 
 ######## URLs ########
 easyrsaVer="3.1.0"
-easyrsaRel="https://github.com/OpenVPN/easy-rsa/releases/download/v${easyrsaVer}/EasyRSA-${easyrsaVer}.tgz"
+easyrsaRel="https://github.com/OpenVPN/easy-rsa"
+easyrsaRel="${easyrsaRel}/releases/download/v${easyrsaVer}"
+easyrsaRel="${easyrsaRel}/EasyRSA-${easyrsaVer}.tgz"
```
orazioedoardo, line 58: Same, as above, it is harder to parse mentally.

auto_install/install.sh
```diff
@@ -62,2762 +69,3686 @@ pivpnPERSISTENTKEEPALIVE=""
...
+  if [[ "${WIREGUARD_BUILTIN}" -eq 1 && -n "${AVAILABLE_WIREGUARD}" ]] \
+    || [[ "${WIREGUARD_BUILTIN}" -eq 1 && ("${PLAT}" == 'Debian' || "${PLAT}" == 'Raspbian') ]] \
+    || [[ "${PLAT}" == 'Raspbian' ]] \
+    || [[ "${PLAT}" == 'Alpine' && ! -f /.dockerenv && "$(uname -mrs)" =~ ^Linux\ +[0-9\.\-]+\-((lts)|(virt))\ +.*$ ]] \
+    || [[ "${PLAT}" == 'Alpine' && -f /.dockerenv ]] \
+    || [[ "${PLAT}" == 'Debian' && "${DPKG_ARCH}" == 'amd64' ]] \
+    || [[ "${PLAT}" == 'Ubuntu' && "${DPKG_ARCH}" == 'amd64' && -n "${AVAILABLE_WIREGUARD}" ]] \
+    || [[ "${PLAT}" == 'Ubuntu' && "${DPKG_ARCH}" == 'arm64' && "${OSCN}" == 'focal' && -n "${AVAILABLE_WIREGUARD}" ]]; then
+    WIREGUARD_SUPPORT=1
+  fi
```
orazioedoardo, line 684: IMHO the description right before the case was easier to read.

auto_install/install.sh
```diff
@@ -62,2762 +69,3686 @@ pivpnPERSISTENTKEEPALIVE=""
...
+    if ! down_dir="$(mktemp -d)"; then
+      err "::: Failed to create download directory for grepcidr!"
+      exit 1
+    fi
+
+    ## download binaries
+    curl \
+      -f \
+      -L \
+      -o "${down_dir}/master.tar.gz" \
```
orazioedoardo, line 728: A new line per option seems overkill

orazioedoardo:
The 80 character limited suggested in the Google style guide is actually too restrictive IMHO, makes too many statements, variable definitions, commands more difficult to read, and probably slower too in cases like [this](https://github.com/pivpn/pivpn/pull/1572/files#diff-1d5b707db02b2c2e0e195e35d2c564c5e044377bb71fac218b315bb3981f856fR42-R66) when you need to call the command many times.

4s3ti:
> The 80 character limited suggested in the Google style guide is actually too restrictive IMHO, makes too many statements, variable definitions, commands more difficult to read, and probably slower too in cases like [this](https://github.com/pivpn/pivpn/pull/1572/files#diff-1d5b707db02b2c2e0e195e35d2c564c5e044377bb71fac218b315bb3981f856fR42-R66) when you need to call the command many times.

Kinda agree, however having too big lines its also not that good. Guess we can agree in being flexible with that rule, and apply the 80 Chars rule whenever possible but open some exceptions where it actually doesn't help much.

giulio-coa:
> Kinda agree, however having too big lines its also not that good. Guess we can agree in being flexible with that rule, and apply the 80 Chars rule whenever possible but open some exceptions where it actually doesn't help much.

For sure, one exception will be " not apply the limit if it make the code less readable and/or harder to parse mentally"

giulio-coa:
@orazioedoardo could you approve and merge the request ?

orazioedoardo:
@giulio-coa I'm not sure I want to, as it is, considering [this](https://github.com/pivpn/pivpn/pull/1572#issuecomment-1209092944). To add, as I said before, [this](https://github.com/pivpn/pivpn/pull/1572/files#diff-72c239a72654ca02e835e59af5796ff45057d3f8b9ba8653f6e64dda7761e2f7R3404-R3416) is much less readable than [this](https://github.com/pivpn/pivpn/pull/1572/files#diff-72c239a72654ca02e835e59af5796ff45057d3f8b9ba8653f6e64dda7761e2f7L2589-L2601).

4s3ti (changes requested):

4s3ti:
Ok then, let's try to not get lost here and not let this drag for too long. 

We should apply the 80 Line rule  whenever it is possible, however let's not get too strict about it and be flexible when to use it or not. 

Let's get this merged when changes are applied, and later one if we feel that we should fix some lines, it can come on separate MR.

4s3ti (approved):

PR #1576, opened 2022-07-27
Remove dependencies

giulio-coa:
Removed the dependency from edge repository and downloaded the `grepcidr` source code into a temp directory

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -585,33 +585,35 @@ preconfigurePackages(){
 	fi
 
 	if [[ "${PLAT}" == 'Alpine' ]] && ! command -v grepcidr &> /dev/null; then
+		local down_dir
 		## install dependencies
 		# shellcheck disable=SC2086
 		${SUDO} ${PKG_INSTALL} build-base make curl tar
 
-		## download binaeries
-		curl -fLo master.tar.gz https://github.com/pivpn/grepcidr/archive/master.tar.gz
-		tar -xzf master.tar.gz
-
-		cd grepcidr-master || exit 1
-
-		## personalize binaries
-		sed -i -E -e 's/^PREFIX\=.*/PREFIX\=\/usr\nCC\=gcc/' Makefile
-
-		## install
-		make
-		${SUDO} make install
-
-		if ! command -v grepcidr &> /dev/null; then
-			echo "::: Failed to compile and install grepcidr!"
+		if ! down_dir="$(mktemp -d)"; then
+			echo "::: Failed to create download directory for grepcidr!"
 			exit 1
 		fi
 
-		cd ..
+		## download binaries
+		curl -fLo "${down_dir}/master.tar.gz" https://github.com/pivpn/grepcidr/archive/master.tar.gz
+		tar -xzC "${down_dir}" -f "${down_dir}/master.tar.gz"
 
-		## remove useless files
-		rm master.tar.gz
-		rm -rf grepcidr-master
+		(
+			cd "${down_dir}/grepcidr-master" || exit 1
+
+			## personalize binaries
+			sed -i -E -e 's/^PREFIX\=.*/PREFIX\=\/usr\nCC\=gcc/' Makefile
+
+			## install
+			make
+			${SUDO} make install
+
+			if ! command -v grepcidr &> /dev/null; then
+				echo "::: Failed to compile and install grepcidr!"
+				exit 1
+			fi
+		) || exit 1
 	fi
 
 	echo "USING_UFW=${USING_UFW}" >> ${tempsetupVarsFile}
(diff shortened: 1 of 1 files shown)


auto_install/install.sh
```diff
@@ -585,13 +585,19 @@ preconfigurePackages(){
...
+			exit 1
+		fi
+
 		## download binaeries
-		curl -fLo master.tar.gz https://github.com/pivpn/grepcidr/archive/master.tar.gz
-		tar -xzf master.tar.gz
+		curl -fLo "${down_dir}/master.tar.gz" https://github.com/pivpn/grepcidr/archive/master.tar.gz
+		tar -xzf "${down_dir}/master.tar.gz"
 
 		cd grepcidr-master || exit 1
```
orazioedoardo, line 602: Missing `$down_dir`

auto_install/install.sh
```diff
@@ -610,7 +616,7 @@ preconfigurePackages(){
 		cd ..
 
 		## remove useless files
-		rm master.tar.gz
+		rm "${down_dir}/master.tar.gz"
 		rm -rf grepcidr-master
```
orazioedoardo, line 620: Here too.

giulio-coa:
Probabilmente hai guardato un vecchio commit

giulio-coa:
You've probably looked at an old commit
These are problems already fixed

orazioedoardo:
> You've probably looked at an old commit

If you go to the _Files changed_ tab you can see that the script is still downloading to the relative directory.

auto_install/install.sh
giulio-coa: ?

orazioedoardo:
> The problems were the downloads of the file not the cd into the directories

It is. Works for grepcidr because you are in the directory where the script is executed, most likely `/home/pi` (although I recommend using absolute paths there too in case someone changes the code above in the future), but it doesn't for `apk-autoupdate` because at that point you are in the `/usr/local/src/pivpn` directory.

`./install.sh: line 2729: cd: apk-autoupdate-master: No such file or directory`

orazioedoardo:
Actually `cd "${down_dir}/apk-autoupdate-master"` still won't work because unzip and tar will extract in the current directory so the script should `cd "${down_dir}"` before unzip/unzip untar first.

auto_install/install.sh
```diff
@@ -2706,13 +2708,34 @@ confUnattendedUpgrades(){
...
+		${SUDO} ${PKG_INSTALL} unzip
 
-		PIVPN_DEPS=(apk-autoupdate)
+		if ! down_dir="$(mktemp -d)"; then
+			echo "::: Failed to create download directory for apk-autoupdate!"
+			exit 1
+		fi
 
-		installDependentPackages PIVPN_DEPS[@]
+		## download binaeries
```
4s3ti, line 2721: typo! :p 

this one just jumped to my eyes (i am usually not good at getting typos, and do a lot of them myself! :p ): binaries

4s3ti:
Following the angular convention, I think with PiVPN the `build` does not apply even tho we're dealing with an "external dependency", we don't rly have a build system. Here I think should be either refactor(): or fix():

orazioedoardo:
Seems to me working but what's up with that `asciidoctor`?

```
[...]
install -d /etc/apk/
install -m 640 etc/autoupdate.conf /etc/apk/
install -d /usr/share/apk-autoupdate
for file in functions.sh openrc.sh; do \
	install -m 644 src/$file /usr/share/apk-autoupdate/$file; \
done
sed -e 's|@VERSION@|0.0.0|g' \
       -e 's|@datadir@|/usr/share/apk-autoupdate|g' \
       -e 's|@sysconfdir@|/etc|g' \
       src/apk-autoupdate.in > build/apk-autoupdate
chmod +x build/apk-autoupdate
cc  -Os -DNDEBUG -std=c11 -DVERSION=0.0.0 -o build/procs-need-restart.o -c src/procs-need-restart.c
cc  -o build/procs-need-restart build/procs-need-restart.o
sed -e 's|@VERSION@|0.0.0|g' \
       -e 's|@datadir@|/usr/share/apk-autoupdate|g' \
       -e 's|@sysconfdir@|/etc|g' \
       src/rc-service-pid.in > build/rc-service-pid
chmod +x build/rc-service-pid
install -d /usr/sbin
for file in apk-autoupdate procs-need-restart rc-service-pid; do \
	install -m 755 build/$file /usr/sbin/$file; \
done
asciidoctor -b manpage -o build/apk-autoupdate.1 man/apk-autoupdate.1.adoc
make: asciidoctor: No such file or directory
make: *** [Makefile:104: build/apk-autoupdate.1] Error 127
rm build/procs-need-restart.o
apk-autoupdate: Checking available updates...
apk-autoupdate: Upgrading packages:  busybox ssl_client busybox-suid python3 openssh-keygen openssh-client-common openssh-client-default openssh-sftp-server openssh-server-common openssh-server openssh
[...]
```

orazioedoardo:
I see that you are not removing downloaded files and build artifacts after compile, intentional?

auto_install/install.sh
```diff
@@ -585,33 +585,35 @@ preconfigurePackages(){
...
+			${SUDO} make install
 
-		## remove useless files
-		rm master.tar.gz
-		rm -rf grepcidr-master
+			if ! command -v grepcidr &> /dev/null; then
+				echo "::: Failed to compile and install grepcidr!"
+				exit 1
+			fi
+		) || exit 1
```
orazioedoardo, line 616: Why wrapping with `(...) || exit 1`? Seems unnecessary considering the inner `exit 1`.

giulio-coa:
> I see that you are not removing anymore downloaded files and build artifacts after compile, intentional? 

Yep
We use a tmp directory, so ...

auto_install/install.sh
orazioedoardo: > The inner exit exits only from the subshell

It does that because it’s wrapped in (…), but it did exit the script before.

orazioedoardo:
> Probably, the dependency is missing
https://asciidoctor.org/

It’s used just for man pages, can you check whether `sudo make install-conf install-data install-exec` suppresses the error? See [here](https://github.com/jirutka/apk-autoupdate/blob/1704b284cbb2a8dd8a7791d13d9741cdfa062aad/Makefile#L53).

auto_install/install.sh
giulio-coa: Yes, but now I don't have to `cd` a subdirectory of my current directory anymore
I have to enter into a directory with a different path and through this trick I avoid using specific Bash functions (in case one day we want to switch completely to the POSIX Shell) and saving the path of the directory I am in and then returning to it (see the piece of code below)
```
local base_dir

base_dir="${PWD}"

cd "${down_dir}/grepcidr-master" || exit 1

...

cd "${base_dir}" || exit 1
```

giulio-coa:
> It’s used just for man pages, can you check whether `sudo make install-conf install-data install-exec` suppresses the error? See [here](https://github.com/jirutka/apk-autoupdate/blob/1704b284cbb2a8dd8a7791d13d9741cdfa062aad/Makefile#L53).

I do, but wouldn't it be better to install `asciidoctor`?
I think it is better to install the package `man` pages

giulio-coa:
> > Probably, the dependency is missing
> > https://asciidoctor.org/
> 
> It’s used just for man pages, can you check whether `sudo make install-conf install-data install-exec` suppresses the error? See [here](https://github.com/jirutka/apk-autoupdate/blob/1704b284cbb2a8dd8a7791d13d9741cdfa062aad/Makefile#L53).

It works

orazioedoardo:
> I do, but wouldn't it be better to install asciidoctor?
I think it is better to install the package's man pages

How much effort/code it is?

giulio-coa:
> How much effort/code it is? I mean, it's not essential.

For Pi-VPN, yes
You should consider that we install on the user's OS a package that he does not know building the source code
It seems appropriate to install the man pages, since it will not be able to install them via `apk`

giulio-coa:
@orazioedoardo So, I can add `asciidoctor` as dependency or we merge the PR as is it ?

orazioedoardo:
> I can add asciidoctor as dependency

Yes

PR #1567, opened 2022-07-16
Added Alpine Linux support

giulio-coa:
Good version of #1566 

I have added the support for Alpine Linux.
The pre-requisites to run Pi-VPN on Alpine are:
* have `bash` installed (`apk add bash`)
* have `busybox-initscripts` installed (`apk add busybox-initscripts`)
* Obviously, run the scripts as `root` or have `sudo` installed (`apk add sudo; echo '%wheel ALL=(ALL) ALL' > /etc/sudoers.d/wheel`)

PR #1580, opened 2022-07-30
Add package dependencies for the install script.

yanxke:
Add package dependencies needed for the install script, or else the script will fail on Linux distributions without these already installed.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -35,7 +35,7 @@ PKG_INSTALL="${PKG_MANAGER} --yes --no-install-recommends install"
 PKG_COUNT="${PKG_MANAGER} -s -o Debug::NoLocking=true upgrade | grep -c ^Inst || true"
 
 # Dependencies that are required by the script, regardless of the VPN protocol chosen
-BASE_DEPS=(git tar curl grep dnsutils grepcidr whiptail net-tools bsdmainutils bash-completion iproute2 ca-certificates)
+BASE_DEPS=(git tar curl grep dnsutils grepcidr whiptail net-tools bsdmainutils bash-completion iproute2 ca-certificates systemctl)
 
 # Dependencies that where actually installed by the script. For example if the script requires
 # grep and dnsutils but dnsutils is already installed, we save grep here. This way when uninstalling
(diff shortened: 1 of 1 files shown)


4s3ti:
we do not intend to support distributions where those are not installed: https://docs.pivpn.io/faq/#what-boardsoses-does-pivpn-support and we are currently in the process of adding support for Alpine (already available on test branch).

 https://github.com/pivpn/pivpn/pull/1567

orazioedoardo:
> Those packages are available by default on the distributions we support.

Maybe there are some VPS out there without those packages (although not having systemd is unusual). Is that the case @yanxke?

4s3ti:
> the script will fail on Linux distributions without these already installed.

I think there are some hard to die debian based distros with the old sysvinit. all those packages come default in any modern distro. haven't found any without it even less on VPS.

yanxke:
> > Those packages are available by default on the distributions we support.
> 
> Maybe there are some VPS out there without those packages (although not having systemd is unusual). Is that the case @yanxke?

It's actually for development and development in Docker on the default Ubuntu image.  Since the scripts look for files in hard coded locations, I don't want to change and mess up my regular settings.  Adding these dependencies don't hurt the normal workflow since they are required anyway.

orazioedoardo:
> Adding these dependencies don't hurt the normal workflow since they are required anyway.

The first two for sure, but the third is not, and even conflicts with other packages (I tested on ubuntu).
```
$ sudo apt install systemctl
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 init : PreDepends: systemd-sysv but it is not going to be installed
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
```
> It's actually for development and debugging in Docker on the default Ubuntu image.

I personally use QEMU with `-snapshot` so that changes are discarded when turning off the VM.

4s3ti:
> > > Those packages are available by default on the distributions we support.
> > 
> > 
> > Maybe there are some VPS out there without those packages (although not having systemd is unusual). Is that the case @yanxke?
> 
> It's actually for development and debugging in Docker on the default Ubuntu image. Since the scripts look for files in hard coded locations, I don't want to change and mess up my regular settings. Adding these dependencies don't hurt the normal workflow since they are required anyway.

for docker is not that linear as adding the dependencies, its actually much more than that and you might be better of with the Alpine support added recently on the test branch. 

you will still have the issue with iptables rules, `net.ipv4.ip_forward` and a bunch more of others. 

PiVPN its not that docker friendly (yet)

PR #1633, opened 2022-10-01
fix(dev-infra): move function before use

rwall:
move err function above where it is first called

### DIFF
--- a/scripts/backup.sh
+++ b/scripts/backup.sh
@@ -6,6 +6,10 @@ screen_size="$(stty size 2> /dev/null || echo 24 80)"
 rows="$(echo "${screen_size}" | awk '{print $1}')"
 columns="$(echo "${screen_size}" | awk '{print $2}')"
 
+err() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $*" >&2
+}
+
 # Divide by two so the dialogs take up half of the screen, which looks nice.
 r=$((rows / 2))
 c=$((columns / 2))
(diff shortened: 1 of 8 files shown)


4s3ti:
Can you please state the reason of your approach?

rwall:
It does fix a bug.
If you trigger an err call before the err function is declared you get a bash error.

Here is what happens if I don't have any clients configured (clients.txt doesn't exist yet):
```
root@adintelmitm28-vm:/usr/local/src/pivpn/scripts/wireguard# pivpn -l
/opt/pivpn/wireguard/listCONF.sh: line 6: err: command not found
```

Here I have the same condition but I moved the function declaration to the top:
```
root@adintelmitm28-vm:/usr/local/src/pivpn/scripts/wireguard# pivpn -l
[2022-10-03T05:57:21+0000]: ::: There are no clients to list
```

4s3ti:
you are right and I understand why ... this is actually due wrong implementation of code style: 

From: https://google.github.io/styleguide/shellguide.html#function-location

> Function Location
>
>Put all functions together in the file just below constants. Don’t hide executable code between functions. Doing so makes the >code difficult to follow and results in nasty surprises when debugging.
>
>If you’ve got functions, put them all together near the top of the file. Only includes, set statements and setting constants may be >done before declaring functions.


and in some cases functions are not actually at the top. 

and it should be: 

Variable declarations
Funcions
other logic

Approving your PR, will deal with that later at some point! 

Thanks a lot for following the code and commit style and taking the time following our guidelines!

4s3ti:
created new issue related with comment above: https://github.com/pivpn/pivpn/issues/1636

PR #1642, opened 2022-10-19
fix(dev-infra): move source to before var ref

rwall:
move file sourcing above first references to the variables it loads

userGroup var is being set using var $install_user that should be loaded from the setupVars.conf file.
the issue this brings is the chown of the configs dir (approx line 100 in makeCONF.sh) leaves that dir owned by root


4s3ti:
Thanks a lot, I Might have missed that one when refactored. 

Please change target branch to test.

4s3ti:
Merged, Thanks a lot :)

PR #1681, opened 2023-01-27
fix(scripts): allow generating a QR code from a filename

austin987:
The last couple times I've used the qr generation functionality, I make the mistake of passing the filename instead of the client name. This time it annoyed me enough to do something about it :).

Thanks for this great tool!


austin987:
Sorry, wrong branch

PR #1682, opened 2023-01-27
  fix(scripts): allow generating a QR code from a filename

austin987:
(supersedes #1681, submitted to the wrong branch, sorry about that).

The last couple times I've used the qr generation functionality, I make the mistake of passing the filename instead of the client name. This time it annoyed me enough to do something about it :).

Thanks for this great tool!


4s3ti:
Hi thanks a lot for your Pull Request, sorry it has taken so long, I genuinely though I had taken care of it. 

this feels more like a feature to me rather than a fix. 
Would you mind to amend your commit header from `fix` to `feat`? other than that all looks good.

austin987:
No worries. Thanks for the feedback. The commit message has been updated.

4s3ti:
Merged. Thanks a lot for your contribution.

PR #1694, opened 2023-02-24
Added Manjaro (Arch) Linux (Tested OpenVPN only) installation capabil…

hammadrauf:
…ity.

'sed' command will edit the respective config files with the fix for the 2 problems: 1) Solve Problem - /var/log permission issue. 2) Solve Problem - iptables-restore rules from file before starting    openvpn service.

Solve /var/log permissions overwrite, Solve iptables-restore execution

Check IP Address settings

Fixing Openvpn startup issues in ManjaroLinux

--

Tested on ManjaroLinux on x86_64/amd64 architecture only. Please feel free to edit as needed. Thanks.


hammadrauf:
Will submit in the test branch.

PR #1696, opened 2023-02-24
feat(platform-server): add Manjaro linux deployment capability

hammadrauf:
Add Manjaro linux (Archlinux) deployment. PiVPN
currently does not support installation on Manjaro. This feature has been tested on Manjaro running
on x86_64/amd64 machine.

Footer.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -44,6 +44,12 @@ BASE_DEPS_ALPINE=(git grep bind-tools newt net-tools bash-completion coreutils)
 BASE_DEPS_ALPINE+=(openssl util-linux openrc iptables ip6tables coreutils sed)
 BASE_DEPS_ALPINE+=(perl)
 
+BASE_DEPS_ARCH=(git tar curl grep bind libnewt net-tools bash-completion)
+BASE_DEPS_ARCH+=(base-devel yay)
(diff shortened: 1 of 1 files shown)


4s3ti:
Thank you for submitting your pull request and the work you have done on it. Unfortunately, I cannot accept it for a few reasons. You could have asked here or searched the repository to find that this topic has been covered before.

1. Arch Linux is a great distribution and has been my main operating system for over 10 years. However, Manjaro/Arch is not a distribution meant for servers. While users can choose to run their servers on any distribution they prefer, it is up to me to decide which distributions I want to support on pivpn. Supporting a new distribution, especially Arch, requires a lot of extra work that we cannot commit to. We would need pipelines to keep testing it and make sure nothing is broken. It would require constant support and testing as Arch packages change often and often introduce breaking changes that require tweaking. I would rather see this effort applied to proper Docker support.
2. Relying on AUR for packages is not something we can do for projects such as pivpn. While AUR is amazing, it is not a source we can rely on. It is not uncommon to have breaking changes there.

Once again, I appreciate the time and effort you put into this, but it does not align with the project goals.

hammadrauf:
No problem. Thanks for letting me know.

Regards

Hammad Rauf

PR #1700, opened 2023-03-02
Allow specifying a third & fourth Wireguard DNS server

cap10morgan:
I wanted to add a couple of IPv6 DNS servers too, but found that it only looked for two. This now also looks for `pivpnDNS3` and `pivpnDNS4` in `setupVars.conf`.

### DIFF
--- a/scripts/wireguard/makeCONF.sh
+++ b/scripts/wireguard/makeCONF.sh
@@ -4,6 +4,8 @@
 # Some vars that might be empty but need to be defined for checks
 pivpnPERSISTENTKEEPALIVE=""
 pivpnDNS2=""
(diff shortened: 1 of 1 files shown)


4s3ti:
Thank you for your pull request, however this PR is quite incomplete I am not going to accept it for now.

This are the main reasons: 

* Please read: https://github.com/pivpn/pivpn#pull-requests 
* DNS are defined during setup time, therefore this feature should added end to end not only when creating configuration files. 
    * At minimum, should be an unattended setup feature. 
* This kind of feature should go on both VPN protocols not only on one


Feel free to re-open it, or open a new one when all those details are sorted out and I will be more than glad to review it again.

PR #1723, opened 2023-05-07
Update install.sh to support wider Ubuntu detection

thadguidry:
Add Ubuntu Lunar (Lobster) which is now available on cloud providers.
Tested with Lunar Lobster 23.04.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -350,7 +350,9 @@ distroCheck() {
       ["11"]="bullseye"
       ["18.04"]="bionic"
       ["20.04"]="focal"
-      ["22.04"]="jammy")
+      ["22.04"]="jammy"
+      ["22.10"]="kinetic"
+      ["23.04"]="lunar")
     OSCN="${VER_MAP["${VER}"]}"
 
     # Alpine support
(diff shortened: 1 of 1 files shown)


orazioedoardo:
It's good to know that the script works but I don't think we want to add non LTS releases. Doing so would mean more commitment to test those releases. The user can still install by clicking through a warning.

4s3ti:
why was this closed?  its not yet available on the ci environment tho.

thadguidry:
@4s3ti Thanks for guidance.  However I am finding it IMPOSSIBLE to amend previous commit messages.  Since this will likely be squashed, can you assist?  I am not even sure what categories this applies to for your CI or documentation.  Is it a feature? Is this change something else?  I leave it to your team to best decide.

4s3ti:
Thank you for your contribution. 
I will add the CI tests as soon as travisci has it available.

about git ... 
I only use git with CLI so i don't know hpw you do it in other tools.

in this partircular case i would use `git reset <commitID>` to go to back to that particular commit without losing any changes, and then commit again with the correct message and force push.

thadguidry:
@4s3ti Ah, not an amend actually then:  `git commit --amend`
, but instead a `git reset`.
When you said "update the commit message" that implied to me to amend.  Since I had already pushed, the `--amend` of course was impossible, so only a `git reset` (undo a commit) could pull me out of the weeds.  Gotcha.

Issue #1733, opened 2023-06-08
[General Issue]: pivpn -off user disables also user-1, user-2...user-n


dolorosus:
### In raising this issue I confirm that

- [X] I have read the [documentation](https://docs.pivpn.io)
- [X] I have read and understood the [PiVPN General Guidelines](https://github.com/pivpn/pivpn/#general-guidelines)
- [X] I have read and understood the [PiVPN Troubleshooting Guidelines](https://docs.pivpn.io/faq/#how-do-i-troubleshoot-connection-issues)
- [X] The issue I am reporting isn't a duplicate, see [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed) and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen).
- [X] I have searched for similar issues and solutions
- [X] I can replicate the issue even after a clean OS installation

### Describe the issue

```pivpn -off user``` disables also user-1, user-2...user-n

```
oot@RASPI4~_#pivpn list
::: Clients Summary :::
Client      Public key                                        Creation date
user        Xt1S..................................tqnUI=      07 Jun 2023, 14:58, CEST
user-1      /wkr..................................5mFnA=      07 Jun 2023, 14:58, CEST
user-2      ajlD..................................NA0Ts=      07 Jun 2023, 14:58, CEST
user-3      J6Ts..................................bAJXM=      07 Jun 2023, 14:58, CEST
::: Disabled clients :::

root@RASPI4~_#pivpn -off user
Confirm you want to disable user? [Y/n] y
user
::: Updated server config
::: Successfully disabled user
::: WireGuard reloaded 
```

Now I expect "user" to be disabled, but user-1, user-2,user-3 have been also disabled

``` 
root@RASPI4~_#pivpn list
::: Clients Summary :::
Client      Public key                                        Creation date
user        Xt1S...................................qnUI=      07 Jun 2023, 14:58, CEST
user-1      /wkr...................................mFnA=      07 Jun 2023, 14:58, CEST
user-2      ajlD...................................A0Ts=      07 Jun 2023, 14:58, CEST
user-3      J6Ts...................................AJXM=      07 Jun 2023, 14:58, CEST
::: Disabled clients :::
[disabled]   user
[disabled]   user-1
[disabled]   user-2
[disabled]   user-3
```

### Expected behavior

[...]

PR #1734, opened 2023-06-09
fix common https://github.com/pivpn/pivpn/issues/1733#issue-1747767321 

dolorosus:
```pivpn -off/-on user``` (en-/dis-)ables also "user-1", "user-2",...,"user-n"
this fix ensures that only "user" is disabled


4s3ti:
Thank you for your pull request.
Merged, 
Squashed commits and redacted the header to match the guidelines.

PR #1744, opened 2023-06-27
Updated README.md

NajmAjmal:
Added extra ! mark


4s3ti:
Curious .... Why an extra mark? what difference does it really make or what exactly does it mean?  

Sure i could google it, but more interested in your take :D

4s3ti:
Merged anyway as to me it doesn't really matter, but I'm still curious about it :)

NajmAjmal:
In informal writing, multiple exclamation points are sometimes used to indicate **stronger emphasis or emotion.**

4s3ti:
That does not answer my question. You changed from 2  !! to !!! what difference does it make??

PR #1750, opened 2023-07-18
Add Alpine container support (Wireguard)

rothdennis:
Running pivpn on a Alpine container under Proxmox was not able to install Wireguard only OpenVPN.


4s3ti:
Hi thank you for your PR however, please read: https://github.com/pivpn/pivpn#pull-requests 

exactly what does this tries to fix? 
how have you tested it? 

also have you tested on regular docker containers or have you tested this on regular docker containers?
As far as I know proxmox uses LXC containers and they work a bit different and this change might break the regular docker images.

rothdennis:
Hi
Thanks for the info. I will close this pull requests and create a new one according to the provided link.

PR #1772, opened 2023-10-20
fix(scripts): Match full client name when disabling

camdenAR:
This fix prevents "pivpn -off" from failing to disable a client when another client is already disabled and the disabled client name starts with the client name you're trying to disable. e.g. "user" wasn't able to be disabled if "user-1" was already disabled.


4s3ti:
Thank you for your contribution :)

PR #1779, opened 2023-11-07
Fix: Change wording on the window

kokomo123:
For grammatical purposes, would this make sense to include wireguard in the window instead of [user] will hold your ovpn configurations. If not, feel free to close this PR as I always wondered why it wasn't added to the window to include wireguard. Thank you!


4s3ti:
In that case I think would make more concise something like:  

"Will hold your VPN client configuration files"

kokomo123:
> In that case I think would make more concise something like:
> 
> "Will hold your VPN client configuration files"

Fair, pushing a fix now! 👍🏻

kokomo123:
Tis done.

4s3ti:
and and merges to master

Issue #1683, opened 2023-01-31
Maximum limit of 254 Wireguard clients


AR1111100011:
I have configured a PiVPN server with Wireguard. The network range I use is pivpnNET=172.31.96.0 and subnetClass=22.
I have reached the client with IP 172.31.97.254. From this point on I can't create any more clients. I have been looking at the makeCONF.sh script and I have seen that it checks the last octet of the clients file.
Is there any way I can continue using clients using the network 172.31.96.0? If I change the network in setupVars.conf I understand that I will still get the same thing, since it checks only the last octet of that file.

orazioedoardo:
Yeah this is something that should be changed. In the meanwhile you can edit `/etc/wireguard/wg0.conf`

AR1111100011:
And what do I get by editing that? The script to generate new users will still not work.

orazioedoardo:
You are right, actually. Are you asking for a feature to change the subnet to a bigger one, or use a bigger subnet for everyone?

AR1111100011:
Currently my network is as follows:

Network: 172.31.96.0/22       	
HostMin:	172.31.96.1	
HostMax:	172.31.99.254	
Broadcast: 172.31.99.255	
Hosts/Net: 1022	

The script started assigning IPs for clients from 172.31.96.2 to 172.31.96.254. The script should still be able to assign IPs, the next one would be the 172.31.97.1

kratkotut:
Have you found a solution? Tell me what you can do.

4s3ti:
Pre-archive closing, more information [here](https://github.com/pivpn/pivpn/releases/tag/v4.6.0)

Issue #1518, opened 2022-05-04
[BUG] Install script hangs generating subnet when physical interface range is 10.0.0.0/8


orazioedoardo:
### Discussed in https://github.com/pivpn/pivpn/discussions/1516

<div type='discussions-op-text'>

<sup>Originally posted by **TimeForANinja** May  3, 2022</sup>
## In raising this issue, I confirm the following:

- [x] I have read the [documentation](https://docs.pivpn.io)
- [ ] Is it a feature request? please consider opening a [Discussion] (https://github.com/pivpn/pivpn/discussions/new)
- [x] I have read and understood the [contributors guide](https://github.com/pivpn/pivpn/blob/master/CONTRIBUTING.md).
- [x] The issue I am reporting can be *replicated*.  
- [x] The issue I am reporting *is* directly related to the pivpn installer script.
- [ ] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pivpn/pivpn/wiki/FAQ), [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed), and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen)).

### Has your install failed? 
```
yes
```

#### Issue
During Installation on my VM the script got stuck after choosing the VPN Type.
CPU usage was at around 100% for ~10 Minutes befor i stopped the script.
Retrying didn't fix it.

#### Investigation
After doing some debugging it looks like the script got stuck inside the [`generateRandomSubnet` Function](https://github.com/pivpn/pivpn/blob/master/auto_install/install.sh#L1214-L1237).
I'm not good with bash scripting but from what i do understand about it it looks like it gets stuck in the `while true;` loop since the subnet of my vm `10.0.0.0/8` will result in all generated pivpnNET `10.XXX.XXX.0` to be invalid...

#### Workaround
I went ahead and edited the script, replacing the `generateRandomSubnet` Function with sth like `echo "10.0.0.1"` - after the installation i edited `/etc/wireguard/wg0.conf`, `/etc/pivpn/wireguard/setupVars.conf` and `/etc/iptables/rules.v4` to my liking.
From what I read the unattended mode can also bypass the function.

#### Possible Fixes
since changing the subnet of your host is not always easily possible i see three options:
1) consider other private ip-ranges in the function
2) manually filter out this edge-case
[...]

PR #1782, opened 2023-11-22
Updates to subnet generation and client creation

orazioedoardo:
First off, now any subnet is allowed and when used, client IP addresses are correctly generated and do not assume /24, fixes https://github.com/pivpn/pivpn/issues/1267.

Then, the script will prevent adding more client than the subnet allows for, fixes https://github.com/pivpn/pivpn/issues/1683.

Last, edge cases in subnet generation when a user has 10.0.0.0/8 is addresses by checking another private network after 5000 iterations. Unfortunately I could not devise a way to efficiently check all empty available subnets so it's still probabilistic. In practice it work fine unless in extreme cases where there are very few holes and those are not checked.  Fixes: https://github.com/pivpn/pivpn/issues/1518.

IPv6 is taken care of by simply using the IPv4 as least significant quartets of the IPv6 address.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -1746,19 +1746,22 @@ dotIPv4LastDec(){
 }
 
 decIPv4ToHex(){
-  local hex ip
-  hex="$(printf "%x\n" "$1")"
-  quartet_hi=${hex::$((${#hex} - 4))}
-  quartet_lo=${hex: -4:4}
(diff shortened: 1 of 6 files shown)


4s3ti:
Will try to take a look at it over the weekend, but you are more than welcome to merge it yourself if I don't :D

orazioedoardo:
Merged so that people can try the test branch

PR #1825, opened 2024-03-19
feat(install) Add possibility to use Pi-hole in unattended install

JonnyMe:
In unattended install, there is no possibility to specify if user wants to use Pi-hole DNS as DNS.
Introducing a `--use-pihole` argument, user can decide if PiVPN configuration should be aligned to the Pi-hole installation.

Example of installation:
```
bash auto_install/install.sh --use-pihole
```
Example of unattended isntallation:
```
bash auto_install/install.sh --unattended pivpn.conf --use-pihole
```

If I made some mistake in the pull request procedure, code style or commit format please, forgive me :-)


4s3ti:
ty!

Issue #1838, opened 2024-05-26
[General Issue]: no packets reaching rasp5 pivpn wireguard


1980Yo:
### In raising this issue I confirm that

- [X] I have read the [documentation](https://docs.pivpn.io)
- [X] I have read and understood the [PiVPN General Guidelines](https://github.com/pivpn/pivpn/#general-guidelines)
- [X] I have read and understood the [PiVPN Troubleshooting Guidelines](https://docs.pivpn.io/faq/#how-do-i-troubleshoot-connection-issues)
- [X] The issue I am reporting isn't a duplicate, see [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed) and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen).
- [X] I have searched for similar issues and solutions
- [X] I can replicate the issue even after a clean OS installation

### Describe the issue

Hi everyone. 
I've been using wireguard vpn for a month or two with no problems at all. Today, I wanted to connect to my server at home and there's no traffic so I tried the steps described here: https://docs.pivpn.io/faq/

### Expected behavior

I cannot get internet access after connecting to my server using Wireguard

### Please describe the steps to replicate the issue

1. Installed Wireguard some months ago with duckdns.org DDNS following instructions and have been running it with no issues whatsover. 
2. Wanted to connect today (via iPhone, via iPad, via PC) with no luck.
3. Tried to solve the problem following instructions found in docs.pivpn.io/faq:
          - $ pivpn -d: selfcheck everything checked out "ok"
          - $ sudo systemctl restart wg-quick@wg0
          - $ lsmod | grep wireguard: 
                           wireguard             114688  0
                           libchacha20poly1305    49152  1 wireguard
                           ip6_udp_tunnel         49152  1 wireguard
                           udp_tunnel             49152  1 wireguard
                           libcurve25519_generic    65536  1 wireguard
                           ipv6                  589824  60 wireguard
          - cat /etc/pivpn/wireguard/setupVars.conf
                           IPv4dev=eth0
                           IPv4addr=192.168.1.109/24
                           IPv4gw=192.168.1.1
                           pivpnPORT=51220
                           pivpnHOST=REDACTED
                           pivpnPROTO=udp
          - ip address matches ("dig REDACTED +short" & "curl -s https://checkip.amazonaws.com" commands used)
         

[...]

PR #1840, opened 2024-06-04
feat(install): add support for wireguard in containers

t0xicCode:
Adds support for detecting when a host kernel loaded wireguard. The kernel module appears loaded but the information about it may not be available.


orazioedoardo:
Thanks, fixes https://github.com/pivpn/pivpn/issues/1838?

t0xicCode:
Not sure that it does.
Instead, this fixes an installation issue where the WireGuard module is
installed on the host and loaded in the kernel, but pivpn is getting
installed in an lxc container (might work for a docker/pod man container
too). Because it’s a container, the `.ko` can be absent and `modinfo
wireguard` can fail, but the WireGuard module is still present and loaded
in the kernel.

Issue #1773, opened 2023-10-22
[General Issue]: Alpine: No routing after reboot (with FIX?)


notreallybob:
### In raising this issue I confirm that

- [X] I have read the [documentation](https://docs.pivpn.io)
- [X] I have read and understood the [PiVPN General Guidelines](https://github.com/pivpn/pivpn/#general-guidelines)
- [X] I have read and understood the [PiVPN Troubleshooting Guidelines](https://docs.pivpn.io/faq/#how-do-i-troubleshoot-connection-issues)
- [X] The issue I am reporting isn't a duplicate, see [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed) and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen).
- [X] I have searched for similar issues and solutions
- [X] I can replicate the issue even after a clean OS installation

### Describe the issue

On Alpine Linux, the iptables is not saved during pivpn install, so after a reboot, clients are not able to reach other servers.

### Expected behavior

iptables should be persisted.

### Please describe the steps to replicate the issue

1. Tested on fresh install of Alpine (under proxmox)
2. Install using curl -L https://install.pivpn.io | bash
3. Run iptables -t nat -L
4. Reboot
5. Run iptables -t nat -L

### Have you taken any steps towards solving your issue?

Only for Alpine:
1. The iptables+ip6tables package doesn't appear to enable the respective services for persistence. This needs to be enabled.
2. Call to enable persistence, similar to existing code under debian/ubuntu.

I modified the auto_install/install.sh script:
1. After package installation, around line 128
  if [[ "${PLAT}" == 'Alpine' ]]; then
    installDependentPackages BASE_DEPS_ALPINE[@]
    ${SUDO} rc-update add iptables    # enable persistence for ipv4
    ${SUDO} rc-update add ip6tables  # enable persistence for ipv6
  else  
    installDependentPackages BASE_DEPS[@]
  fi

2. Near line 3391, add case to persist.
  case "${PLAT}" in   
    Debian | Raspbian | Ubuntu)
      ${SUDO} iptables-save \         
        | ${SUDO} tee /etc/iptables/rules.v4 > /dev/null
      ${SUDO} ip6tables-save \    
        | ${SUDO} tee /etc/iptables/rules.v6 > /dev/null
      ;;                                                             
[...]

PR #1849, opened 2024-07-29
Fix Alpine persistence

wlmchen:
Alpine saves iptables rules differently than Debian, and the respective services need to be enabled to persist settings across reboots. The only related issue I could find is #1773.


orazioedoardo:
Thanks

PR #1899, opened 2025-03-10
feat(install): Remove support for EOL OS

coolapso:
Debian buster is EOL since June 30th 2024


orazioedoardo:
Good to see you @coolapso.

PR #1900, opened 2025-03-21
(fix): Fix IPv6 address in config file

dot-1q:
I found this "error" (in quotes because it doesn't really break anything) while configuring WGDashboard. [Issue Here](https://github.com/donaldzou/WGDashboard/issues/658)

Essentially, the `wg0.conf` file layout changed and the IPv4 and IPv6 addresses are in different lines in the `[Interface]` section, as opposed to being separated by a comma.

You can check this by having a Wireguard setup with IPv4 and IPv6 and then run `wg-quick save wg0`. You'll see that the `wg-quick` utility will format the file with one line per address.

This PR helps bring uniformity with the `wg-quick` utility.

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -3270,10 +3270,10 @@ confWireGuard() {
   {
     echo '[Interface]'
     echo "PrivateKey = $(${SUDO} cat /etc/wireguard/keys/server_priv)"
-    echo -n "Address = ${vpnGw}/${subnetClass}"
+    echo "Address = ${vpnGw}/${subnetClass}"
 
     if [[ "${pivpnenableipv6}" -eq 1 ]]; then
-      echo ",${vpnGwv6}/${subnetClassv6}"
+      echo "Address = ${vpnGwv6}/${subnetClassv6}"
     else
       echo
     fi


dot-1q:
To be honest, this seems to me more like a Wireguard mistake, since TOML doesn't really [support duplicate keys](https://toml.io/en/v1.0.0#keyvalue-pair), and wireguard config file seems to borrow from its syntax. 

Wonder if its best to make a PR against the wg-quick utility 

[config file](https://git.zx2c4.com/wireguard-tools/tree/src/wg-quick/linux.bash#n257) generation.

DaanSelen:
> To be honest, this seems to me more like a Wireguard mistake, since TOML doesn't really [support duplicate keys](https://toml.io/en/v1.0.0#keyvalue-pair), and wireguard config file seems to borrow from its syntax.
> 
> Wonder if its best to make a PR against the wg-quick utility, though they use the same format for every OS
> 
> [config file](https://git.zx2c4.com/wireguard-tools/tree/src/wg-quick/linux.bash#n257) generation.

Any help is welcome, and since wg-quick is a "script" if I am not mistaken we can also manually change it, but for that we should know what we are doing.

PR #1914, opened 2025-07-02
Fixed routing

hmodi51:
This PR adds a safe optional call to **'netfilter-persistent save'** after the NAT MASQUERADE rule is added for openVPN users. 

Why?

Currently, PiVPN adds the NAT MASQUERADE rule for OpenVPN, and also saves iptables rules using `iptables-save` on Debian/Ubuntu/Raspbian. However, in certain environments, this save step may not catch the NAT rule due to timing, platform detection issues, or early reboot.
This change adds a fallback `netfilter-persistent save` to ensure that iptables rules including the NAT rule are persistent across reboots. 

What this does :

- Adds a conditional check if netfilter-persistent is present and if it does then calls **'netfilter-persistent save'** which ensures iptables rules are saved explicitly after NAT setup.

This does not affect WireGuard users, nor does it alter existing rule logic. It's a minimal reliability improvement that addresses edge case where NAT rules are sometimes lost after reboot.

Let me know if you prefer it to be refactored

### DIFF
--- a/auto_install/install.sh
+++ b/auto_install/install.sh
@@ -823,7 +823,7 @@ installDependentPackages() {
 
   if [[ "${FAILED}" -gt 0 ]]; then
     echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]:" >&2
-    ${SUDO} cat "${APTLOGFIGitHub accountLE}" >&2
+    ${SUDO} cat "${APTLOGFILE}" >&2
     exit 1
   fi
 }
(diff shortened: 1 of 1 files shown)


hmodi51:
This might not seem like an important change but I had to scratch my brain the whole day as I previously had setup my vpn using pivpn but maybe that time I skipped through the installation of iptables-persistent or maybe something else happened so its just a fallback method if it was installed but skipped saving

orazioedoardo:
Doesn't `netfilter-persistent save` run `iptables-save > /etc/iptables/rules.v4` anyway?

PR #1922, opened 2025-09-27
Update install.sh

Vesrol:
Generate a different IPv6 ULA on each installation to avoid collisions when a peer connects to multiple instances.


orazioedoardo:
Seems a bit confusing at the end to use the same low bits of the gateway address as the higher bits of the prefix.

PR #1926, opened 2025-11-06
feat(install): Remove support for EOL OS

mwoolweaver:
Updated version mapping for Debian and Ubuntu releases with support expiration dates as comments.


orazioedoardo:
Debian bullseye should still be partially supported as LTS. It might make more sense to drop support for Ubuntu focal or both since that requires paid ESM support. Obviously those still work anyway.

coolapso (commented):
howdy .. just passing by, I really don't like to see this AI-like inline comments, also they feel meaningless, they are relevant only for anyone maintaining the codebase or contributing and anyone doing so should know where to find that information anyway! 

just my 2 cents.

PR #1929, opened 2025-12-27
fix(install): use mktemp to prevent TOCTOU race condition

cooldadhacking:
## Summary

Local privesc via TOCTOU race on `/tmp/setupVars.conf`. Attacker wins race between `rm` and first write, creates world-writable file. Bash `>` preserves ownership, so attacker can poison config before copy to `/etc/pivpn/`. All pivpn scripts source this as root.

## Fix

Use `mktemp` (same pattern as lines 736, 799, 3784 in this file).

## Testing

Verified race no longer exploitable on Debian with WireGuard install.


orazioedoardo:
Thanks, the pull request should first target the test branch though.

cooldadhacking:
Np! I changed the target

Issue #1935, opened 2026-03-26
[General Issue]: no systemctl on Alpine Linux


thenktor:
### In raising this issue I confirm that

- [x] I have read the [documentation](https://docs.pivpn.io)
- [x] I have read and understood the [PiVPN General Guidelines](https://github.com/pivpn/pivpn/#general-guidelines)
- [x] I have read and understood the [PiVPN Troubleshooting Guidelines](https://docs.pivpn.io/faq/#how-do-i-troubleshoot-connection-issues)
- [x] The issue I am reporting isn't a duplicate, see [closed issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aclosed) and [open issues](https://github.com/pivpn/pivpn/issues?q=is%3Aissue+sort%3Aupdated-desc+is%3Aopen).
- [x] I have searched for similar issues and solutions
- [x] I can replicate the issue even after a clean OS installation

### Describe the issue

Hi,

just installed pivpn on Alpine Linux 3.23.3. When creating a new client config an error about systemctl appears, which of course does not exist on Alpine:

```txt
pivpn add -n testconf
grep: warning: stray \ before =
Enter the Client IP from range 10.142.77.2 - 10.142.77.254 (optional):
::: Chosen Client IP: 10.142.77.2
::: Client Keys generated
::: Client config generated
::: Updated server config
/opt/pivpn/wireguard/makeCONF.sh: line 291: systemctl: command not found
[2026-03-26T11:43:53+0100]: ::: Failed to reload WireGuard
======================================================================
::: Done! testconf.conf successfully created!
::: testconf.conf was copied to /home/pivpn/configs for easytransfer.
::: Please use this profile only on one device and create additional
::: profiles for other devices. You can also use pivpn -qr
::: to generate a QR Code you can scan with the mobile app.
======================================================================
```

### Expected behavior

systemctl should not be used on Alpine Linux.

### Please describe the steps to replicate the issue

Run `pivpn add -n testconf` on Alpine Linux.

### Have you taken any steps towards solving your issue?

no

### Where did you run pivpn?

Alpine Linux 3.23.3

### Please provide your output from `uname -a`

Linux example.com 6.18.19-0-lts #1-Alpine SMP PREEMPT_DYNAMIC 2026-03-20 15:33:57 x86_64 GNU/Linux

### Details about Operative System

NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.23.3
[...]

PR #1936, opened 2026-03-26
Fix for recent grep versions and Alpine check

thenktor:
Fixes an error on Alpine Linux caused by grep 3.12 when using unnecessary backslash. See #1935


orazioedoardo:
You should target the test branch, not master.

thenktor:
Sorry, changed it.

PR #1942, opened 2026-07-16
Fix debug script to show more than one client

LeePorte:
I spotted that the script would only ever return the first client. It now returns each one.


orazioedoardo:
It’s intentional to only show one which is enough since other clients should only differ by the last octet of the IP and also avoids sending too much text when opening an issue

